PR #15, opened 2019-06-06
Support dynamic imports for babel transformation

### LINKED ISSUE #14: Support dynamic imports in Jest testing
Using dynamic imports in web applications is a common theme to minimize initial download time for web clients, like here with webpack:

`const d3 = await import(/* webpackChunkName: "d3js" */ 'd3');`

When running Jest in `--coverage` mode Jest throws the error `Support for the experimental syntax 'dynamicImport' isn't currently enabled`.

muenzpraeger:
Fixes #14


packages/@lwc/jest-transformer/src/index.js
```diff
@@ -30,6 +31,7 @@ const BABEL_CONFIG = {
     presets: [jestPreset],
     plugins: [
         babelCommonJs,
+        babelSyntaxDynamicImport,
```
trevor-bliss:
```suggestion
        babelDynamicImport,
```

muenzpraeger: Fixed

muenzpraeger:
Any chance that we can get this in 1.1?

caridy:
I'm good with this, but I wonder how does this work for the resolver logic? e.g.:

```js
import Foo from 'x/foo';
const Bar = import('x/bar');
```

assuming that foo and bar are components, does it resolves correctly?

caridy:
sure, I understand that, but there is the ergonomic aspect of this, which is, not predictable. Some folks will use it for libs, and it will work great, but some others will attempt to use it for other imports... and it will not work, e.g.: a local library from the same namespace.

muenzpraeger:
I think @pmdartus uses it already without issues.

https://github.com/pmdartus/rcast/blob/14aa560d09dbefd86ab04d3b5ef7c4a96c41ede2/src/client/modules/rcast/app/app.js#L32

caridy:
@muenzpraeger that's precisely the part that I want to understand. @pmdartus how does that work? will the dynamic import be transformed during compilar (resolved) that eventually the babel plugin will work fine to transform that into a regular require? What happen if we provide a string value computed dynamically? `import(someComputedValue)`?

muenzpraeger:
You can see the compiled output here:

http://lwc-recipes-oss-pr-9.herokuapp.com/

I’m printing the output in the console (boo, I know).

pmdartus:
Currently, LWC compiler completely ignores the `import()`. When rollup or webpack receives the compiled source code from the LWC compiler, their own resolution `import()` resolution kicks off. 

When passing a string literal, webpack and rollup will automatically do the code splitting for you by creating multiple chunks and renaming the reference to point to the right chunk.
The story is different for computed properties since it can't be resolved by the bundler the developers needs to manually create the different chunks by hand and make sure the computed property points to one of his chunks.

IMO, the LWC compiler should not be in the business of transforming the `import()`. Transforming the `import()` requires the knowledge off all the modules present in the application. The consumer of the lwc compiler should be in charge of figuring that out:
* webpack / rollup: use the tool's capabilities to transform the `import()` syntax and genrate the appropriate chunks
* node / jest: transform the `import()` to a commonjs `require()`
* Salesforce platform: custom module loading tailored for the platform 

I think it worth migrating this discussion to the main [salesforce/lwc](https://github.com/salesforce/lwc) repo if we want to discuss how to handle dynamic import.

caridy:
@pmdartus that doesn't explain how it works today for 

```
import Foo from 'x/foo';
const Bar = import('x/bar');
```

If LWC compiler is doing nothing for `x/bar`, how does it work?

pmdartus:
For example in the case of rollup, when returning the following code `import('x/bar');`. Rollup parses the code and invokes the `resolveId` rollup hook with the `'x/bar'` identifier. Like any other module, we delegate the resolution to the lwc module resolver. 
The LWC compiler is not involved in the module resolution.

muenzpraeger:
Where did we land on this now?

muenzpraeger:
Thoughts?

muenzpraeger:
@trevor-bliss can we get this is with the next release?

diervo:
I'm fine merging this, however if we update to the latests babel this won't be necessary since its already supported by default.

trevor-bliss:
@muenzpraeger would you mind updating this PR and resolving the conflicts? Then we can merge it in.

muenzpraeger:
Done, @trevor-bliss. Thanks!

aputinski:
@trevor-bliss Any idea when this fix will be released?

trevor-bliss:
@aputinski I can do a release today.

trevor-bliss:
@muenzpraeger @aputinski This is released under `2.1.0` now.

muenzpraeger:
@aputinski You'll have to use a version prior to `1.3.0 beta` of `lwc-services` for that.

PR #17, opened 2019-06-11
refactor: starting lwc 1.1.x, jest tests will run native shadow

caridy:
Related to https://github.com/salesforce/lwc/pull/1322

Synthetic shadow, starting in v1.1.0, will not longer run in jest/jsdom since that supports native shadow.


caridy:
@trevor-bliss @diervo how do you want to release this? Diego has some ideas on how to support both options depending on the version of the engine... but that's going to be tricky.

trevor-bliss:
Why not just a peer dependency on the preset for @lwc/engine >= 1.1.0? I'm skeptical to change behavior based on the engine version we detect while running tests.

caridy:
Actually, I think it is safe since 1.0 should work without synthetic as well, just need to have the jsdom@15 on it. I think it is safe to merge this and release it, not need to fork logic based on the version of the engine.

PR #33, opened 2019-09-04
wdio-lwc-service 1.1.0

Templarian:
This need to be published as a `beta`.

`npm publish --tag beta`. The version can be 1.0.0, but this may not be the last beta if anything comes up or changes are made to `lwc@1.1` or `lwc-create-app`.


Templarian:
Going to find some time shortly to work on on this again. Busy with internal work at the moment.

albert-sf:
#40

gonzalocordero:
What's the general goal of this? Add wdio support? We should drive an RFC if there is not any for this. I can help with that.

Do we have a set of requirements/story?

Templarian:
@gonzalocordero An RFC would be a great way of handling this tbh. I think we were kind of adhoc doing this before to shim a performant solution for 1.0. If you want to schedule some time I can step through what this service does. Realistically I won't have time to actually put code into a solution, but can outline the goals in the RFC and how the tests "sandboxes" need to be setup for WDIO to consume the tests.

The goal of this service will be the same, but the actual code could probably be cleaned up now that we're on 1.5 and not using a lot of internal methods.

ekashida:
@Templarian we recently moved wdio-lwc-service outside of the logical monorepo because of the overhead it was incurring on our release process (i.e., manual interventions between lerna commands to avoid releasing new versions) as well as the security alert originating from its dependency tree which we didn't have the time to address. I will close this PR since it looks like there are considerable conflicts now. Feel free to either reopen this one or create a new one with updated changes.

Issue #35, opened 2019-09-05
Jest: Support of TypeScript for LWC Open Source


muenzpraeger:
When using TypeScript for LWC Open Source projects Jest is failing. This is because of a missing Babel pre-transform step, before the LWC compiler transform the source. 

See the discussion [here](https://github.com/muenzpraeger/create-lwc-app/issues/92#issuecomment-528232156).

Templarian:
@muenzpraeger Did this go out in 1.3?

muenzpraeger:
Yes and no, @templarian.

I am using my own forks till @tbliss publishes an LWC 1.1 build that contains that (now that LWC `latest` is on 1.1).

trevor-bliss:
The PR to consume LWC 1.1 is here: https://github.com/salesforce/lwc-test/pull/26

PR #36, opened 2019-09-05
Adding TypeScript support for LWC OSS

muenzpraeger:
This PR adds the necessary changes to pre-transform TypeScript files (see #35).

It fixes the issue opened here:

https://github.com/muenzpraeger/create-lwc-app/issues/92

### DIFF
--- a/packages/@lwc/jest-transformer/src/index.js
+++ b/packages/@lwc/jest-transformer/src/index.js
@@ -13,8 +13,8 @@ const lwcCompiler = require('@lwc/compiler');
 const jestPreset = require('babel-preset-jest');
 const babelTsPreset = require.resolve('@babel/preset-typescript');
 const babelCommonJs = require('@babel/plugin-transform-modules-commonjs');
-const babelClassProperties = require('@babel/plugin-syntax-class-properties');
-const babelSyntaxDecorators = require('@babel/plugin-syntax-decorators');
+const babelClassProperties = require('@babel/plugin-proposal-class-properties');
+const babelSyntaxDecorators = require('@babel/plugin-proposal-decorators');
 
 const engineVersion = getVersion();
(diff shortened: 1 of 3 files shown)


packages/@lwc/jest-transformer/src/index.js
```diff
@@ -11,7 +11,10 @@ const { getVersion } = require('lwc');
 const babelCore = require('@babel/core');
 const lwcCompiler = require('@lwc/compiler');
 const jestPreset = require('babel-preset-jest');
+const babelTsPreset = require.resolve('@babel/preset-typescript');
 const babelCommonJs = require('@babel/plugin-transform-modules-commonjs');
+const babelClassProperties = require('@babel/plugin-syntax-class-properties');
```
trevor-bliss: Is there a reason you used this package instead of `@babel/plugin-proposal-class-properties`? The message at the top of the doc for this one seems to recommend `@babel/plugin-proposal-class-properties`.

https://babeljs.io/docs/en/babel-plugin-syntax-class-properties

packages/@lwc/jest-transformer/src/index.js
```diff
@@ -11,7 +11,10 @@ const { getVersion } = require('lwc');
...
+const babelSyntaxDecorators = require('@babel/plugin-syntax-decorators');
```
trevor-bliss: Same comment as above. Doc seems to imply `@babel/plugin-proposal-decorators` is better to use.

packages/@lwc/jest-transformer/src/index.js
muenzpraeger: Good catch, updated.

packages/@lwc/jest-transformer/src/index.js
muenzpraeger: Good catch, updated.

PR #41, opened 2019-10-07
Add support for LWC modules from npm packages in lwc 1.1

muenzpraeger:
The current module resolution doesn't take into account modules that are imported via NPM.

As they need to be passed explicitly the dependencies of the current project are passed to the module resolver, similar as with `create-lwc-app`.

FYI @diervo


Templarian:
Will this handle someone placing their utils in a sub package? Haven't dived into all the resolver updates in `resolveModules` yet.

`foo-project` <- `foo-base-components` <- `foo-utils`

midzelis:
@trevor-bliss Can you please merge this when you get a chance.

PR #42, opened 2019-10-22
chore: lwc 1.1.x support

ekashida:
Jest tests now use jsdom's shadow implementation instead of `@lwc/synthetic-shadow`.


packages/wdio-lwc-service/examples/basic/src/index.html
```diff
@@ -10,6 +10,10 @@
 <body>
     <div id="main"></div>
 </body>
+    <script>
+        typeof process === 'undefined' && (process = { env: { NODE_ENV: 'development' } });
+    </script>
+    <script src="/synthetic-shadow.js"></script>
```
caridy: do wd really needs synthetic shadow?

ekashida: This is for setting up integration tests so I believe we do need it.

trevor-bliss:
@ekashida I assume this is supposed to replace https://github.com/salesforce/lwc-test/pull/26? 

If so, we should make sure https://github.com/salesforce/lwc-test/pull/26/commits/e038c8e5c8f223e3d8922482ed8ce7c96bcd4668 does not get forgotten about since that was merged directly into Jose's PR assuming that would make it into `master`.

packages/wdio-lwc-service/examples/basic/src/index.html
trevor-bliss: We want customers to be able to use the wdio service for all our "supported" browsers so I think it makes sense for the example to include synthetic shadow. Customers can remove it if they don't need it.

ekashida:
@trevor-bliss https://github.com/salesforce/lwc-test/issues/43

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -31,7 +31,7 @@ const WHITELISTED_LWC_PACKAGES = {
     'wire-service': '@lwc/wire-service',
     'wire-service-jest-util': 'lwc-wire-service-jest-util',
 };
-const lwcMap = lwcNpmResolver.resolveLwcNpmModules();
+const lwcMap = lwcResolver.resolveModules();
```
tedconn: The API has changed and it now returns an array. This breaks the resolver.

ekashida:
I messed up and published the package in my "alpha" branch before updating/merging this PR so I ended up pushing the change to master as f798d5082bfcfed8f3d6bc1b680df5fc8906c4a9.

package.json
```diff
@@ -14,16 +14,14 @@
...
-    "@lwc/synthetic-shadow": "~1.0.1",
-    "@lwc/wire-service": "~1.0.1",
+    "@lwc/compiler": "^1.1.5",
+    "@lwc/module-resolver": "^1.1.5",
     "babel-eslint": "~10.0.1",
     "eslint": "^5.16.0",
     "eslint-plugin-webdriverio": "^1.0.1",
     "jest": "~24.8.0",
     "lerna": "~3.14.1",
+    "lwc": "^1.1.5",
```
tedconn: lwc already includes the @lwc/compiler

PR #51, opened 2019-12-09
chore: remove unmaintained example package

ekashida:
We can direct users to https://github.com/salesforce/lwc-todomvc for an example of how to use wdio with lwc.


ekashida:
@Gr8Gatsby and I had a quick meeting with @garazi about this and he said he would get back to us after discussing further with @reiniergs and @Templarian.

trevor-bliss:
I need more context here. What is the original motivation for this change?

The PR title vs. files changed feels a little misaligned to me so just to make sure we're on the same page, here is my initial thoughts:

The PR title refers to removing the "example package", but I see the entire `wdio-lwc-service` package is being removed. `wdio-lwc-service` is more than an example of how to use wdio to test LWC. It's a wdio [custom service](https://webdriver.io/docs/customservices.html) with the goal of enabling testing of LWC components in isolation. The `lwc-todomvc` package is a good example of using wdio, but requires loading the full app for every test rather than testing individual components. 

The package is also open sourced and I've seen it used internally by other teams at Salesforce. As @pmdartus mentioned above, if we do decide to remove it I think we need a more thorough deprecation and communication strategy at the very least.

Templarian:
Yeah, `wdio-lwc-service` does a lot of extra things that ensure that tests run much faster by creating / running component tests relatively isolated way.

https://github.com/salesforce/lwc-docs/pull/72 It's also tied to the documentation here that we put on hold due to the 1.1 updating breaking compatibility of the service.

ekashida:
The current discussion is around potentially moving the package somewhere else since it would be maintained by Base Components.

ekashida:
https://github.com/salesforce/lwc-test/pull/83

PR #57, opened 2020-01-09
build: fix incorrect dependencies

ekashida:
Looks like I screwed up some of the dependencies while working in `lwc-test`. This PR fixes those and also fixes some existing dependency misconfigurations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "license": "MIT",
   "devDependencies": {
     "@lwc/compiler": "1.1.13-224.4",
+    "@lwc/errors": "1.1.13-224.4",
     "@lwc/synthetic-shadow": "1.1.13-224.4",
     "babel-eslint": "~10.0.1",
     "eslint": "^5.16.0",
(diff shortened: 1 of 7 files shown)


packages/@lwc/jest-transformer/src/index.js
```diff
@@ -75,7 +73,7 @@ module.exports = {
             .createHash('md5')
             .update(JSON.stringify(options), 'utf8')
             .update(
-                fileData + filePath + configStr + NODE_ENV + compilerVersion + engineVersion,
```
ekashida: Removed because I assumed that jest tests are unlikely to have mixed versions.

package.json
```diff
@@ -14,15 +14,15 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@lwc/compiler": "1.1.13-224.3",
-    "@lwc/module-resolver": "1.1.13-224.3",
+    "@lwc/compiler": "1.1.13-224.4",
+    "@lwc/errors": "1.1.13-224.4",
+    "@lwc/synthetic-shadow": "1.1.13-224.4",
```
ekashida: These are dev deps because some packages in lwc-test use @lwc/jest-preset to configure their jest testsand these are peer deps of @lwc/jest-preset.

PR #64, opened 2020-02-25
build: upgrade dependencies

ekashida:
Also removes wdio dependency and configuration from the root workspace.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@lwc/compiler": "1.1.15",
-    "@lwc/engine": "1.1.15",
-    "@lwc/synthetic-shadow": "1.1.15",
+    "@lwc/compiler": "1.3.2",
+    "@lwc/engine": "1.3.2",
+    "@lwc/synthetic-shadow": "1.3.2",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -14,16 +14,15 @@
...
-    "@lwc/engine": "1.1.15",
-    "@lwc/synthetic-shadow": "1.1.15",
-    "babel-eslint": "~10.0.1",
-    "eslint": "^5.16.0",
+    "@lwc/compiler": "1.3.2",
+    "@lwc/engine": "1.3.2",
+    "@lwc/synthetic-shadow": "1.3.2",
+    "babel-eslint": "~10.0.3",
+    "eslint": "^6.8.0",
     "eslint-plugin-webdriverio": "^1.0.1",
```
pmdartus: This is not needed since you removed it from the plugin section.

ekashida:
@Templarian @reiniergs Just an fyi, I modified wdio deps and configs since they should not have been part of the root workspace. I also disabled esilnt in wdio-lwc-service since I kept getting eslint configuration errors related to `eslint-plugin-webdriverio` and I didn't have time to dig into what the issue was.

PR #65, opened 2020-03-09
feat: add TypeScript support for LWC OSS

### LINKED ISSUE #35: Jest: Support of TypeScript for LWC Open Source

muenzpraeger:
Fixes #35

### DIFF
--- a/packages/@lwc/jest-transformer/src/index.js
+++ b/packages/@lwc/jest-transformer/src/index.js
@@ -11,7 +11,10 @@ const babelCore = require('@babel/core');
 const lwcCompiler = require('@lwc/compiler');
 const jestPreset = require('babel-preset-jest');
 const babelCommonJs = require('@babel/plugin-transform-modules-commonjs');
+const babelClassProperties = require('@babel/plugin-proposal-class-properties');
 const babelDynamicImport = require('babel-plugin-transform-dynamic-import');
+const babelSyntaxDecorators = require('@babel/plugin-proposal-decorators');
+const babelTsPreset = require.resolve('@babel/preset-typescript');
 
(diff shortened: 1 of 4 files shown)


muenzpraeger:
Would be great to get this out soon, so that I can revert back to the original LWC test packages, and get rid of my custom forks. ;-)

packages/@lwc/jest-transformer/src/index.js
```diff
@@ -51,8 +67,24 @@ const BABEL_CONFIG = {
...
+        ...BABEL_TS_CONFIG,
+        filename: filePath,
+    });
+    return code;
+}
+
 module.exports = {
     process(src, filePath) {
+        if (isTypeScript(filePath)) {
+            src = transformTypeScript(src, filePath);
```
jodarove: looks good @muenzpraeger , just one note:

For ts files, you are doing **ts -> "lwc"** -> js -> "jest". In the ts -> "lwc" transformation no sourcemaps is generated. this means that the code coverage is not against the ts file, but against the generated "lwc" code; also you wont be able to debug in the ts file.

That being said, there is no way today to pass an input sourcemap to the LWC compiler (takes care of lwc -> js). An alternative to explore, if you want to have the coverage against the .ts file, is to rebuild the sourcemap with the partial transformations sourcemaps: 
ts -> lwc => map1
lwc -> js => map2
map1 + map2 = map3 (which maps from ts -> js)

PR #67, opened 2020-03-11
fix: add missing TS test match pattern

muenzpraeger:


muenzpraeger:
Can we get this in, and have a new version published? Would be great! @jodarove

jodarove:
done @muenzpraeger , the new version containing your change is 4.5.1

PR #72, opened 2020-03-26
fix: 🐛 change dependencies for TS

muenzpraeger:


muenzpraeger:
Not sure what type of test you recommend, @trevor-bliss.

muenzpraeger:
Slimming down - got it. I duplicated the existing ones, for consistency. Will update.

muenzpraeger:
@trevor-bliss Added simplified test cases.

aputinski:
When is the next release due to be published? And will this be included?

PR #74, opened 2020-04-02
feat: 🎉 implement LWC npm module resolution

### LINKED ISSUE #73: Support for LWC npm module resolution
The current version of LWC Test does not resolve 3rd party modules that are installed via npm.

The new module resolution for LWC has landed in v1.4.0, based on [this RFC](https://rfcs.lwc.dev/rfcs/lwc/0020-module-resolution). To support testing with imported modules the resolver package needs to implement the new mechanism.

muenzpraeger:
Fixes #73


muenzpraeger:
FYI @pozil @diervo

muenzpraeger:
Ok, this will take somewhat longer.

In the previous (non-supported) `module-resolver` implementation of LWC it was much simpler, as you'd only had to drop in the dependencies of a project, and it did the resolution.

diervo:
Wait, I don't think you need any of that. Now all you need to do is read from the config and just let the resolver do its thing, we should not need to do any of the previous. I need to talk to @trevor-bliss to see if we can simplify this.

muenzpraeger:
Does that mean everyone has to maintain a dedicated config of which modules they want to use from a package within their own projects?

diervo:
Well right now you have to manually create a regular expression on jest for mapping don't you? :)

The idea will be to use the same config to reference your tests, but we can debate it to make it ergonomics.

muenzpraeger:
@diervo @trevor-bliss LMK if this works for you know.

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -53,6 +55,19 @@ function getLwcPath(path, options) {
         return require.resolve(WHITELISTED_LWC_PACKAGES[path]);
     }
 
+    // If it could be an LWC module from npm, resolve it relative to this folder
+    if (path.search('^[A-Za-z]+/.*') > -1) {
```
diervo, line 59: Why do we need this RegEx?

muenzpraeger: `path` can be many things, like an npm package name, or a file, or a module. I thought to only invoke the module resolver when path is in the form `abc/def`.

diervo: But do we care? the module will either resolve or throw... Why do we need regex?

muenzpraeger: I thought it would be more efficient (perf wise), as we don't have to call the resolve all the time. Once you've a few dependencies in the project that can be easily 40-50 times. If you tell me it doesn't matter - b/c you're more the perf master than me - I can happily remove it.

diervo: I would rather optimize later, for example scoped modules will not work in your regex (ex @foo/bar)

muenzpraeger: Great argument! Regex removed.

packages/@lwc/jest-resolver/package.json
```diff
@@ -11,11 +11,14 @@
         "/resources/",
         "/src/index.js"
     ],
+    "dependencies": {
+        "@lwc/module-resolver": "^1.5.0"
```
trevor-bliss, line 15: I'm sure this will get out of date unless we make it a peer dependency or we build a script/CI that automatically keeps it up to date.

muenzpraeger:
Yeah, I was similar not that happy to add the dependency. At the same time I think re-building the logic is more expensive.

packages/@lwc/jest-resolver/package.json
diervo: agree, however is unlikely this will change often, so we are mostly fine

ekashida:
Left a comment [here](https://github.com/salesforce/lwc-test/issues/73#issuecomment-623795717) about why we don't think we should merge this change.

Some interesting ideas about introducing a standard way for published components to expose test stubs to dependent components were floated around...I wonder if you think such a feature would help out with the issue you're facing?

muenzpraeger:
@ekashida Got it, and I can work around that. At the end I still think we need to provide some kind of mechanism to allow that.

IMHO the current POV is much driven on how we have the capability on providing all the stubs for our own components. As we want to drive broader adoption I think there's a necessity to have default mechanism, as we can't (and shouldn't expect) from developers to build stubs for 3rd party components that they want to use. No matter how technically correct it is, it'll become in the long run a testing adoption blocker, IMHO.

ekashida:
@muenzpraeger Thanks for adding visibility to this issue. I'll bring it up with the team so that @Gr8Gatsby can help prioritize.

Issue #66, opened 2020-03-10
Typescript sourcemap support


jodarove:
With #65 we added typescript support for LWC opensource.

In #65 a new transformation was added for `.ts` files, and they are transpiled to "js with lwc" syntax, leaving the transformation process for .ts files like: `ts -> "lwc" -> js -> "jest"`. In the ts -> "lwc" transformation no sourcemaps is generated, this results in the following issues:

1. The code coverage is not against the `.ts` file, but against the generated "lwc" code.
2. The debuggable code is the "lwc" generated code on the first transformation, when ideally, it should be the typescript code.

jodarove:
cc/ @muenzpraeger

jodarove:
solved by #75

PR #75, opened 2020-04-08
Typescript sourcemaps

dexwiz:
This addresses issue #66 

Added source maps to Typescript step, and passed them to LWC Compiler.


muenzpraeger:
@jodarove

dexwiz:
@jodarove Looked it it further, and it seems that inline sourcemaps are enough to get sourcemaps to work during debugging.

PR #76, opened 2020-04-17
chore: upgrade to lwc version 1.5.0

jodarove:
With the [Wire protocol reform](https://github.com/salesforce/lwc-rfcs/blob/master/text/0000-wire-reform.md), the wire-service now requires a wire-adapter (even a mock) in the proper format, otherwise the test of a component using a wire will throw during rendering.

This PR hijacks the decorators registration system of the engine to replace the used wire-adapter (maybe invalid) with a valid one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
         "@lwc/compiler": "^1.5.0",
         "@lwc/engine": "^1.5.0",
         "@lwc/synthetic-shadow": "^1.5.0",
+        "@lwc/wire-service": "^1.5.0",
         "@types/jest": "^25.1.4",
         "babel-eslint": "^10.1.0",
         "eslint": "^6.8.0",
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
     ],
     "license": "MIT",
     "devDependencies": {
-        "@lwc/compiler": "^1.3.2",
-        "@lwc/engine": "^1.3.2",
-        "@lwc/synthetic-shadow": "^1.3.2",
+        "@lwc/compiler": "^1.5.0",
+        "@lwc/engine": "^1.5.0",
+        "@lwc/synthetic-shadow": "^1.5.0",
         "@types/jest": "^25.1.4",
         "babel-eslint": "^10.1.0",
         "eslint": "^6.8.0",
--- a/packages/@lwc/jest-preset/src/__tests__/simpleComponent.spec.js
+++ b/packages/@lwc/jest-preset/src/__tests__/simpleComponent.spec.js
@@ -6,67 +6,12 @@
  */
 import { createElement } from 'lwc';
 import SimpleComponent from 'example/simpleComponent';
-import { wireAdapter } from 'example/adapter';
 
 afterEach(() => {
     while (document.body.firstChild) {
         document.body.removeChild(document.body.firstChild);
     }
 });
-
-/**
- * Example method used in wire-service-jest-util
- * @param adapterId
- * @returns {{getLastConfig: (function(): *), emit: emit}}
- */
-function registerTestAdapter(adapterId) {
-    let lastConfig;
-    const wiredEventTargets = [];
-
-    const emit = (value) => {
-        wiredEventTargets.forEach(wiredEventTarget => wiredEventTarget.emit(value));
-    };
-
-    const getLastConfig = () => {
-        return lastConfig;
-    };
-
-    const add = (arr, item) => {
-        const idx = arr.indexOf(item);
-        if (idx === -1) {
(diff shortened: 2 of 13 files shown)


packages/@lwc/jest-preset/src/setup.js
```diff
@@ -23,3 +23,105 @@ if (shadowRootPrototype.$constructorCache$ === undefined) {
...
+
+function overriddenRegisterDecorators(Ctor, decoratorsMeta) {
+    const wire = decoratorsMeta.wire || {};
+
+    Object.keys(wire).forEach((adapterName) => {
+        const adapter = wire[adapterName].adapter;
+
+        if (!global.adapterIdToAdapterMockMap.has(adapter)) {
+            // validate the adapter
+            if (!Object.isExtensible(adapter)) {
```
pmdartus, line 75: It took me a while to understand why this check is needed. It's because we attach a property on `adapter` line 95. It would be better to add a more meaningful comment above: `Checking if the adapter is extensible since we will need to add the adapter property on it.`

packages/@lwc/jest-preset/src/__tests__/simpleComponent.spec.js
```diff
@@ -6,13 +6,68 @@
...
 });
 
+/**
+ * Example method used in wire-service-jest-util
+ * @param adapterId
+ * @returns {{getLastConfig: (function(): *), emit: emit}}
+ */
+function registerTestAdapter(adapterId) {
+    let lastConfig;
+    const wiredEventTargets = [];
```
pmdartus, line 24: Make this a Set instead of an Array. We can use Sets here since this code doesn't run on IE11.

packages/@lwc/jest-preset/src/__tests__/simpleComponent.spec.js
```diff
@@ -6,13 +6,68 @@
...
+    };
+
+    const add = (arr, item) => {
+        const idx = arr.indexOf(item);
+        if (idx === -1) {
+            arr.push(item);
+        }
+    };
+
+    const relatedAdapter = global.adapterIdToAdapterMockMap.get(adapterId);
```
pmdartus, line 41: Let's handle gracefully the case where the `relatedAdapter` is not defined.

packages/@lwc/jest-preset/src/setup.js
```diff
@@ -23,3 +23,105 @@ if (shadowRootPrototype.$constructorCache$ === undefined) {
     global.CustomEvent = CustomEvent;
     global.MutationObserver = MutationObserver;
 }
+
+// Provides temporary backward compatibility for wire-protocol reform: lwc > 1.5.0
+global.adapterIdToAdapterMockMap = new Map();
```
pmdartus, line 28: This is a global property that is exposed to everybody and that will be consumed by another package, so I would like to make it clear it's a temporary hack in the identifier. What do you think about using the `wireAdaptersRegistryHack` or something like that?

packages/@lwc/jest-preset/src/__tests__/simpleComponent.spec.js
jodarove: with https://github.com/salesforce/lwc-test/pull/76/commits/dc8ceefd174f5cd43680058d17d603a6bca9c553 related adapter, will always be defined.

package.json
```diff
@@ -15,9 +15,10 @@
     ],
     "license": "MIT",
     "devDependencies": {
-        "@lwc/compiler": "^1.3.2",
-        "@lwc/engine": "^1.3.2",
-        "@lwc/synthetic-shadow": "^1.3.2",
+        "@lwc/compiler": "^1.5.0",
+        "@lwc/engine": "^1.5.0",
+        "@lwc/synthetic-shadow": "^1.5.0",
+        "@lwc/wire-service": "^1.5.0",
```
jodarove, line 21: needed for testing purposes

packages/@lwc/jest-preset/src/setup.js
```diff
@@ -23,3 +23,152 @@ if (shadowRootPrototype.$constructorCache$ === undefined) {
...
+
+    return newAdapterMock;
+}
+
+function overriddenRegisterDecorators(Ctor, decoratorsMeta) {
+    const wire = decoratorsMeta.wire || {};
+
+    Object.keys(wire).forEach((adapterName) => {
+        const adapter = wire[adapterName].adapter;
+        let wireAdapterMock = global.wireAdaptersRegistryHack.has(adapter);
```
jodarove, line 128: must be a get

PR #77, opened 2020-04-21
fix(jest-preset): registerDecorators proxy does not returns original call result

jodarove:
Fixes an issue with the `registerDecorator` proxy, that does not returns the result from calling the original `lwc.registerDecorator`

### DIFF
--- a/packages/@lwc/jest-preset/src/test/modules/example/wiredComponent/wiredComponent.js
+++ b/packages/@lwc/jest-preset/src/test/modules/example/wiredComponent/wiredComponent.js
@@ -7,13 +7,7 @@
 import { LightningElement, api, wire } from 'lwc';
 import { realAdapter } from 'example/adapter';
 
-function getNavigationMixin(BaseClass) {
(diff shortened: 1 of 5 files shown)


packages/@lwc/jest-preset/src/test/modules/example/wiredComponent/wiredComponent.js
```diff
@@ -7,7 +7,13 @@
 import { LightningElement, api, wire } from 'lwc';
 import { realAdapter } from 'example/adapter';
 
-export default class WiredComponent extends LightningElement {
+function getNavigationMixin(BaseClass) {
```
ravijayaramappa, line 10: Did you intend to make this component extend a mixin? _Asking because it does not relate to the PR title or description_

jodarove: > Did you intend to make this component extend a mixin?
Yes, this implementation triggers the issue.

PR #90, opened 2020-06-13
chore: update peerdependency of jest to ^25.5.4 @W-7689277

ravijayaramappa:


ekashida:
We should probably also make babel-jest-preset a peer dependency as well since this change will require a major version bump.

package.json
```diff
@@ -20,8 +20,9 @@
         "@lwc/wire-service": "^1.5.2",
         "@types/jest": "^25.2.1",
         "babel-eslint": "^10.1.0",
+        "babel-preset-jest": "^25.5.0",
```
ravijayaramappa, line 23: Adding this as a devDependency here as the unit tests in `@lwc/jest-transformer` rely on this dependency.

packages/@lwc/jest-preset/package.json
```diff
@@ -30,7 +30,7 @@
         "@lwc/jest-resolver": "7.0.0",
         "@lwc/jest-serializer": "7.0.0",
         "@lwc/jest-transformer": "7.0.0",
-        "jest-environment-jsdom-fifteen": "^1.0.2"
+        "jest-environment-jsdom-fifteen": "~1.0.2"
```
ravijayaramappa, line 33: Only change is to use the right semver qualifier for dependency. The version in yarn.lock remains the same.

packages/@lwc/jest-transformer/package.json
```diff
@@ -28,17 +28,18 @@
...
-        "jest": ">= 24.9.0"
+        "babel-preset-jest": "^25.5.0",
+        "jest": "^25.5.4"
     },
     "devDependencies": {
         "@lwc/jest-preset": "7.0.0",
         "@lwc/jest-resolver": "7.0.0",
-        "@lwc/jest-serializer": "7.0.0"
+        "@lwc/jest-serializer": "7.0.0",
+        "babel-preset-jest": "^25.5.0"
```
ravijayaramappa, line 42: Adding this as a devDependency here as the unit tests in @lwc/jest-transformer rely on this dependency.

packages/@lwc/jest-preset/package.json
pmdartus, line 33: `jest-environment-jsdom-fifteen` is not needed anymore since `jest` 25 depends on jsdom 15: https://github.com/facebook/jest/blob/v25.5.4/packages/jest-environment-jsdom/package.json#L25

packages/@lwc/jest-transformer/package.json
```diff
@@ -28,17 +28,18 @@
...
         "@babel/preset-typescript": "7.9.0",
         "@babel/template": "~7.8.6",
-        "babel-plugin-transform-dynamic-import": "^2.1.0",
-        "babel-preset-jest": "^25.4.0"
+        "babel-plugin-transform-dynamic-import": "~2.1.0"
     },
     "peerDependencies": {
         "@lwc/compiler": "*",
-        "jest": ">= 24.9.0"
+        "babel-preset-jest": "^25.5.0",
```
pmdartus, line 35: I am not certain if `babel-preset-jest` should be peer dependency. Since `babel-preset-jest` is already pulled down by `jest`, and by using a peer dependency we might run into a version mismatch. What do you think about resolving the `babel-preset-jest` from the `jest` module?

trevor-bliss: When we tested it, the `babel-preset-jest` dependency was in the node_modules of `jest-config` rather than at the top level. We could dig around node_modules to find it. That feels a bit  fragile though, `jest-config` doesn't even have a README. 

Or maybe there's a more reliable way to get the `babel-preset-jest` from `jest`?

packages/@lwc/jest-preset/package.json
```diff
@@ -24,13 +24,12 @@
     "peerDependencies": {
         "@lwc/compiler": "*",
         "@lwc/synthetic-shadow": "*",
-        "jest": ">=24.9.0"
+        "jest": "^25.5.4"
```
ekashida, line 27: Please add `jest` as a peerDependency in `@lwc/jest-transformer`, `@lwc/jest-serializer`, and `@lwc/jest-resolver' as well.

packages/@lwc/jest-preset/jest-preset.js
```diff
@@ -17,8 +17,4 @@ module.exports = {
     // temp workaround until this is released - https://github.com/facebook/jest/pull/6792
     testURL: 'http://localhost/',
     coveragePathIgnorePatterns: ['.css$', '.html$'],
-
-    // Use an up-to-date version of jsdom. Jest v24 comes with jsdom v11 which doesn't offer support
-    // for native shadow DOM.
-    testEnvironment: 'jest-environment-jsdom-fifteen',
```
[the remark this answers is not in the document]
ravijayaramappa: They have been removed in the same commit https://github.com/salesforce/lwc-test/pull/90/commits/4ed55d387754fc2d0d79b56e6e1973528c78fa39

packages/@lwc/jest-resolver/package.json
```diff
@@ -20,9 +20,15 @@
         "/resources/",
         "/src/index.js"
     ],
+    "dependencies": {
+        "jest-resolve": "~25.5.1"
```
ekashida, line 24:
```suggestion
        "jest-resolve": "25.5.1"
```

PR #96, opened 2020-08-26
chore: relax jest peer dependency

ekashida:
We still need to make sure that `jest` is at least v25.5.4 to avoid the `@jest/globals` issue.

### DIFF
--- a/packages/@lwc/jest-preset/package.json
+++ b/packages/@lwc/jest-preset/package.json
@@ -24,7 +24,7 @@
     "peerDependencies": {
         "@lwc/compiler": "*",
         "@lwc/synthetic-shadow": "*",
-        "jest": "^25.5.4"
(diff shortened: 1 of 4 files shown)


packages/@lwc/jest-preset/package.json
```diff
@@ -24,7 +24,7 @@
     "peerDependencies": {
         "@lwc/compiler": "*",
         "@lwc/synthetic-shadow": "*",
-        "jest": "^25.5.4"
+        "jest": ">=25.5.4"
```
pmdartus, line 27: If we relax the jest peer dependency version, I would recommend adding a new test suite to the CI to check against all the supported versions of jest. Here is an example of how it can be done: https://github.com/salesforce/eslint-plugin-lwc/pull/51.

Issue #102, opened 2020-10-12
@lwc/jest-serializer: fails on null value


rochejul:
Hi

When we want to use "toMatchSnapshot" with LWC and Jest, we could have the following error (when some properties of the JSON have got a null value):

```
PrettyFormatPluginError: Cannot destructure property `nodeType` of 'undefined' or 'null'.TypeError: Cannot destructure property `nodeType` of 'undefined' or 'null'.

at Object.test (node_modules/@lwc/jest-serializer/src/index.js:12:14)
      at __EXTERNAL_MATCHER_TRAP__ (node_modules/expect/build/index.js:342:30)
      at Object.expect [as toMatchSnapshot] (node_modules/expect/build/index.js:343:15)
      at it.each (src/main/modules/search_lightning/searchResultsAuraAdapter/__tests__/modelMapper.test.js:88:13)
```

When we look on the code, we have:

```js
function test({ nodeType } = {}) {
    return (
        nodeType &&
        (nodeType === 1 || // element
        nodeType === 3 || // text
            nodeType === 6) // comment
    );
}
```

According to ES6, the default value is applied only if the parameter is undefined. If we have a null value, it will not be applied.

Then, we should made instead:

```
function test(node) {
    const nodeType = node && node.nodeType || undefined;
    return (
        nodeType &&
        (nodeType === 1 || // element
        nodeType === 3 || // text
            nodeType === 6) // comment
    );
}
```

Should I create a PR?

Many thanks

Regards

pmdartus:
Thanks for reporting the issue @julien-reboul.
Feel free to open a PR if you have the code ready to go.

julien-reboul:
> Thanks for reporting the issue @julien-reboul.
> Feel free to open a PR if you have the code ready to go.

@pmdartus you meant @rochejul 😄

pmdartus:
Fixed by #103.

PR #103, opened 2020-10-13
fix(jest-serializer): issue when value is null

rochejul:
PR related to #102 issue

The PR contains a test file for index.js and then tests related to the "test" function.

The test function has evolved to:
* Always return a boolean
* Take care of null value

Here a screenshot of the tests:
Screenshot 2020-10-13 at 09 30 38


.gitignore
```diff
@@ -6,4 +6,5 @@ node_modules/
 coverage/
 .idea/
 .project/
-.vscode/
\ No newline at end of file
+.vscode/
+.history/
```
pmdartus, line 10: Why ignoring `.history` folder?

packages/@lwc/jest-serializer/src/__tests__/index.spec.js
```diff
@@ -0,0 +1,20 @@
+const { test } = require('../index');
+
+describe('@lwc/jest-serializer', () => {
+    describe('and "test" method', () => {
+        const nodeTypeArr = [
+            ['undefined', undefined, false],
+            ['null', null, false],
+            ['empty object', { }, false],
+            ['without the nodeType property', { anotherKindOfProperty: 5 }, false],
+            ['with an element node type', { nodeType: 1 }, true],
```
pmdartus, line 10:
```suggestion
            ['with an element node type', document.createElement('div'), true],
```

packages/@lwc/jest-serializer/src/__tests__/index.spec.js
```diff
@@ -0,0 +1,20 @@
...
+            ['with a text node type', { nodeType: 3 }, true],
```
pmdartus, line 11:
```suggestion
            ['with a text node type', document.createTextNode(''), true],
```

packages/@lwc/jest-serializer/src/__tests__/index.spec.js
```diff
@@ -0,0 +1,20 @@
...
+            ['with a comment node type', { nodeType: 6 }, true],
```
pmdartus, line 12:
```suggestion
            ['with a comment node type', document.createComment(''), true],
```

packages/@lwc/jest-serializer/src/__tests__/index.spec.js
```diff
@@ -0,0 +1,20 @@
...
+            ['with another kind of node type', { nodeType: 2 }, false],
```
pmdartus, line 13:
```suggestion
            ['with another kind of node type', document.createAttribute('foo'), false],
```

packages/@lwc/jest-serializer/src/index.js
```diff
@@ -9,8 +9,9 @@ const DOMElement = PrettyFormat.plugins.DOMElement;
 
 const cleanElementAttributes = require('./clean-element-attrs');
 
-function test({ nodeType } = {}) {
-    return (
+function test(node) {
```
pmdartus, line 12: Let's make this method a little more comprehensive, by checking first if the passed argument is an object and then let's make sure if the object is a DOM node that can be serialized.

```suggestion
function test(obj) {
  if (typeof obj !== 'object' || obj === null) {
    return false;
  }
  
  const { nodeType } = obj;
  return (
        nodeType === 1 || // element
        nodeType === 3 || // text
        nodeType === 6    // comment
    );
}
```

pmdartus (changes requested):
Thanks for the PR, added a couple of comments.

.gitignore
rochejul: There is a VScode plugin for code history which create an ".history" folder. And so to avoid to commit, we should ignore it.

If needed, I could revert it

packages/@lwc/jest-serializer/src/index.js
rochejul: It make senses

packages/@lwc/jest-serializer/src/index.js
```diff
@@ -9,12 +9,18 @@ const DOMElement = PrettyFormat.plugins.DOMElement;
...
+    }
+
+    const { nodeType } = obj;
+
+    // According https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType
+    return !!(
+        nodeType === 1 || // element
         nodeType === 3 || // text
-            nodeType === 6) // comment
+        nodeType === 8    // comment
```
rochejul, line 23: With "document.createComment" (and according to MDN), the comment's nodeType is 8 (6 represents the ENTITY instruction for DTD)

pmdartus: Good catch.

pmdartus (approved):

PR #108, opened 2021-01-12
chore: Project dependencies update

pmdartus:
## Details

- Update all the `peerDependencies` to match
- Update all the dependencies versions and move to a `^` version pinning (more details on this below)
- Remove unused packages: `babel-plugin-transform-dynamic-import`, `@babel/template`
- Normalize `license`, `keywords` and `main` fields between projects

**Why moving to `^` versioning pinning?** 
Historically, we decided that using `~` was the safest approach to version dependencies for LWC related projects to prevent unexpected downstream breaking changes. I do think that this project doesn't need to follow the same semantic:
- Jest uses `^` to pin [dependencies](https://github.com/facebook/jest/blob/v26.6.3/packages/jest-cli/package.json) (regardless if the dependency is external to jest or not). Because of this, even if we protect this project from potential Babel breaking change, Jest doesn't. Using `^` better aligns with Jest dependency management.
- Reduces maintenance burden on us for this less critical project.

### DIFF
--- a/packages/@lwc/jest-transformer/src/index.js
+++ b/packages/@lwc/jest-transformer/src/index.js
@@ -11,6 +11,7 @@ const babelCore = require('@babel/core');
 const lwcCompiler = require('@lwc/compiler');
 const jestPreset = require('babel-preset-jest');
 const babelCommonJs = require('@babel/plugin-transform-modules-commonjs');
+const babelDynamicImport = require('@babel/plugin-proposal-dynamic-import');
 const babelClassProperties = require('@babel/plugin-syntax-class-properties');
 const babelSyntaxDecorators = require('@babel/plugin-syntax-decorators');
 const babelTsPreset = require.resolve('@babel/preset-typescript');
@@ -51,6 +52,7 @@ const BABEL_CONFIG = {
     sourceMaps: 'both',
     presets: [jestPreset],
     plugins: [
+        babelDynamicImport,
         babelCommonJs,
         apexScopedImport,
         apexContinuationScopedImport,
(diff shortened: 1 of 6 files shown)


packages/@lwc/jest-transformer/src/index.js
```diff
@@ -53,7 +52,6 @@ const BABEL_CONFIG = {
     presets: [jestPreset],
     plugins: [
         babelCommonJs,
-        babelDynamicImport,
```
jodarove, line 56: I have doubts about removing this; if `babel-plugin-transform-dynamic-import` is removed, how will components using `import()` work in the jest environment?

jodarove (changes requested):

packages/@lwc/jest-transformer/src/index.js
pmdartus: You are right @jodarove this plugin is still needed. This is something interesting because the commonjs plugin refers in the documentation that it is [capable to transform `import()` expression](https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs). I will open a thread on the Babel repro to discuss this.

jodarove: it could be because of [this check](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-modules-commonjs/src/index.js#L132). That being said, I like more this approach: using `@babel/plugin-proposal-dynamic-import` (+modules-commonjs) instead of `babel-plugin-transform-dynamic-import`.

jodarove (approved):
:shipit:

PR #109, opened 2021-01-13
test: Move tests to end to end approach

pmdartus:
## Details

This PR aims to simplify the test setup of this repository to match as close as possible to how the `@lwc/jest-preset` is actually used by downstream projects.

While working on #108, I realized that this project has a really complex test setup with multiple levels of `jest.config.js`. This is because we try to unit test each project individually. Since those packages are intended to work together, this PR a new internal package to the monorepo called `test`. The soil goal of this package is to exercise the resolver, transform, and preset logic like a standard consumer of the `@lwc/jest-preset` consumer would do.

This approach gets rid of the complex setup and centralizes all the tests under this new internal package.


pmdartus:
Currently I half of the test have been migrated. 
@jodarove @ekashida @ravijayaramappa  Let me know what you think of this new testing approach and I will migrate the rest of the tests.

test/src/modules/serializer/component/component.html
```diff
@@ -0,0 +1,3 @@
+<template>
+    <h1>I an LWC component</h1>
```
trevor-bliss, line 2:
```suggestion
    <h1>I am an LWC component</h1>
```

trevor-bliss:
@pmdartus I think this approach is great. It makes sense since these packages are meant to be used together. The transforms in the jest-transformer package may still have value as standalone unit tests though. I'm not sure we want to convert those.

package.json
```diff
@@ -2,16 +2,18 @@
...
-    "main": "index.js",
     "scripts": {
-        "test": "jest --config ./scripts/jest/root.config.js",
+        "test": "yarn test:unit && yarn test:integration",
+        "test:unit": "jest --config ./scripts/jest/root.config.js --no-cache",
+        "test:integration": "jest --config ./test/jest.config.js --no-cache",
         "clean": "lerna run clean && lerna clean --yes && rm -rf node_modules",
         "lint": "eslint packages/ scripts/",
-        "format": "prettier --write '{packages,scripts}/**/*.{js,ts,json,md}'",
+        "format": "prettier --write '{packages,scripts,test}/**/*.{js,ts,json,md}'",
```
ravijayaramappa, line 11: while you are at it
```suggestion
        "format": "prettier --write '{packages,scripts,test,spec}/**/*.{js,ts,json,md}'",
```

test/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "test",
+    "version": "0.0.0",
```
ravijayaramappa, line 3: nit: optional for private packages
```suggestion
```

test/src/modules/resolver/noHtml/__tests__/noHtml.test.js
```diff
@@ -5,11 +5,8 @@
...
-import Basic from '../x-basic';
+import NoHTML from '../noHtml';
 
-describe('x-basic', () => {
-    it('loads basic component', () => {
-        const element = createElement('x-basic', { is: Basic });
-        expect(element).toBe(element);
-    });
+it('resolves component without HTML', () => {
+    expect(() => createElement('resolver-no-html', { is: NoHTML })).not.toThrow();
```
ravijayaramappa, line 11: Going by the other resolver tests, it seems like you want to test that the component renders. If that is the case: The template is not resolved until the element is connected to the document.

test/src/modules/serializer/component/__tests__/component.spec.js
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { createElement } from 'lwc';
+import Component from '../component';
+
+it('resolves component without HTML', () => {
```
ravijayaramappa, line 10: The test component does have a template.
```suggestion
it('serializes component with HTML', () => {
```

ravijayaramappa (commented):
The new end to end tests are great. :shipit: 
I concur with Trevor's point. There is value in having directly relatable unit tests for each package. May be file an issue to follow up with simplified unit tests.

ekashida (commented):
Sounds like a great idea!

test/package.json
pmdartus: This is in fact needed by `yarn`: 

```
warning Missing version in workspace at "/Users/p.dartus/code/github/salesforce/lwc-test/test", ignoring.
```

package.json
pmdartus: This not needed, all the subpaths are already included in this glob.

pmdartus:
Thanks for the great feedback. The PR is now ready for a final round of review.
I agree with you both @trevor-bliss and @ravijayaramappa. I kept the unit tests in place and added at the root a new `jest.config.json` to split the test suite into 2 buckets: the unit tests (using Jest defaults) and the integration tests (using the LWC jest preset). We can get the best of both worlds this way.

Something interesting to note here is that I am passing `--no-cache` for all the tests. This ensures that Jest always uses the preset code and doesn't relied on previously transformed code. This is important for the integration tests.

package.json
ravijayaramappa: My bad, thought these were file name suffix. Didn’t realize they were folder names 🤦

ravijayaramappa (approved):

trevor-bliss (approved):

PR #110, opened 2021-01-22
fix: add @lwc/engine as peerDependency of @lwc/jest-preset

jodarove:
`@lwc/engine` is not marked as a peer dependency of `@lwc/jest-preset`. However, `@lwc/engine` is required during the [environment setup](https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-preset/src/setup.js#L149).

### DIFF
--- a/packages/@lwc/jest-preset/package.json
+++ b/packages/@lwc/jest-preset/package.json
@@ -22,6 +22,7 @@
     ],
     "peerDependencies": {
         "@lwc/compiler": "*",
+        "@lwc/engine": "*",
         "@lwc/synthetic-shadow": "*",
         "jest": "^26"
     },


trevor-bliss (approved):

packages/@lwc/jest-preset/package.json
```diff
@@ -22,6 +22,7 @@
     ],
     "peerDependencies": {
         "@lwc/compiler": "*",
+        "@lwc/engine": "*",
         "@lwc/synthetic-shadow": "*",
```
ekashida, line 26: Should we specify `lwc` here instead of individual `@lwc/*` packages? We tell users to add `lwc` as a dependency instead of individual `@lwc/*` packages in their projects, but these users see installation warnings when they install our `lwc-test` packages because our `lwc-test` packages list out `@lwc/*` packages as peer dependencies.

```suggestion
        "lwc": "*",
```

ravijayaramappa:
@jodarove Please close the loop on this PR.

pmdartus (approved):

ekashida:
Should we hold off on making this change until after we deprecate `@lwc/engine` in favor of `@lwc/engine-dom`?

PR #113, opened 2021-03-08
fix: allow class declarations from TS->JS

### LINKED ISSUE #111: Typescript components are not transpiled with observable fields
There seems to be a difference in the way javascript components and typescript components are transpiled when running tests via Jest, causing reactive updates to fail as fields are no longer registered.

Looking into the Jest cache we see that the transformed TS generated by @lwc/jest-transform does not contain a constructor initializing the fields and their registerDecorators differ in that the fields property is missing.

```
//Transformed JS component
  constructor(...args) {
    super(...args);
    this._label = void 0;
    this._value = void 0;
    this._width = void 0;
    this._prefix = void 0;
    this._postfix = void 0;
  }

....

(0, _lwc.registerDecorators)(CurrencyInput, {
  publicProps: {
    label: {
      config: 3
    },
    value: {
      config: 3
    },
    width: {
      config: 3
    },
    prefix: {
      config: 3
    },
    postfix: {
      config: 3
    }
  },
  fields: ["_label", "_value", "_width", "_prefix", "_postfix"]
});
```

```
//Transformed TS component
(0, _lwc.registerDecorators)(CurrencyInput, {
  publicProps: {
    label: {
      config: 3
    },
    value: {
      config: 3
    },
    width: {
      config: 3
    },
    prefix: {
      config: 3
    },
    postfix: {
      config: 3
    }
  }
});
```

I've forked the LWC-test repo and reworked the Typescript component/test so it reproduces the issue and made a similiar javascript component for comparison here: https://github.com/JSGTrifork/lwc-test/tree/bug/typescript-test

The end result is that tests must be synchronous, that is that all property assignment must happen before the component is mounted. This limits the types of scenarios we can test in our unit testing.

abdulsattar:
Fixes #111

### DIFF
--- a/packages/@lwc/jest-transformer/src/index.js
+++ b/packages/@lwc/jest-transformer/src/index.js
@@ -46,7 +46,7 @@ const BABEL_TS_CONFIG = {
             },
         ],
     ],
-    presets: [babelTsPreset],
+    presets: [[babelTsPreset, { allowDeclareFields: true }]],
 };
 
 const BABEL_CONFIG = {
--- a/test/src/modules/smoke/typescript/__tests__/__snapshots__/typescript.test.ts.snap
+++ b/test/src/modules/smoke/typescript/__tests__/__snapshots__/typescript.test.ts.snap
@@ -9,5 +9,8 @@ exports[`example-typescript should render correct snapshot 1`] = `
     <p>
       Validating if the transformation to TypeScript has worked.
     </p>
+    <div>
+      Validating if unitialized  is tracked.
+    </div>
 </example-typescript>
 `;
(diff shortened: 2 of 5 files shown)


pmdartus (changes requested):
Could you add a test for this to prevent future regressions?

JSGTrifork:
Any tractions on this? The unit test should be similar in nature to the one from the repro-case in #111, feel free to use it as you please.

JSGTrifork:
@abdulsattar if there is anything we can do to move this forward, please reach out.

pmdartus (approved):

abdulsattar:
Apologies for taking so long @JSGTrifork

PR #115, opened 2021-04-05
chore: circleci npm deploy

ekashida:
Deploy npm packages using circleci


.circleci/config.yml
```diff
@@ -1,30 +1,105 @@
...
+      only: /^v.*/
+
+
+executors:
+  node:
+    working_directory: ~/lwc-test
     docker:
-      - image: circleci/node:lts
-    working_directory: ~/repo
+      - image: circleci/node:14
```
ekashida, line 17: Circleci best practice is to use an exact version number so this is probably a good middle ground.

PR #118, opened 2021-05-25
[Breaking] chore(deps): update lwc dependencies (@W-9318804)

nolanlawson:
Updates to the latest version of lwc-related packages. Switches from the deprecated `@lwc/engine` to `@lwc/engine-dom`.

I don't think this is a breaking change, because `@lwc/engine-dom` is also available in 1.x.

W-9318804


rsalvador (approved):

packages/@lwc/jest-preset/package.json
```diff
@@ -22,7 +22,7 @@
     ],
     "peerDependencies": {
         "@lwc/compiler": "*",
-        "@lwc/engine": "*",
+        "@lwc/engine-dom": "*",
```
ravijayaramappa, line 25: IMO, the release after this PR is merged should be tagged a breaking change because users have to make a change in their dependencies to consume this version.

ravijayaramappa (approved):

packages/@lwc/jest-preset/package.json
nolanlawson: Ah good point, I missed this. This is a breaking change, yes.

PR #119, opened 2021-05-27
fix: update dep versions

nolanlawson:
I made a mistake during the v11.0.0 update, so I skipped some dependency versions, causing tests to fail.

### DIFF
--- a/test/package.json
+++ b/test/package.json
@@ -4,8 +4,5 @@
     "private": "true",
     "scripts": {
         "test": "jest"
-    },
-    "devDependencies": {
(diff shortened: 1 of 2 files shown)


test/package.json
```diff
@@ -4,8 +4,5 @@
     "private": "true",
     "scripts": {
         "test": "jest"
-    },
-    "devDependencies": {
-        "@lwc/jest-preset": "10.0.2"
```
nolanlawson, line 9: I don't see any reason for this to be here; it gets hoisted up into the top-level `node_modules` anyway.

ravijayaramappa (approved):

PR #122, opened 2021-06-08
feat: add option to test native shadow DOM

### LINKED ISSUE #93: @lwc/synthetic-shadow installed even when native Shadow DOM is available
Looks like this check is returning `undefined` even when native Shadow DOM is available (JSDOM 16).

https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-preset/src/setup.js#L9

I'm guessing `$constructorCache$` was changed at some point in JSDOM.

Our LWC components are used with native Shadow DOM, so we have a few tests that check things like `assignedNodes()`

nolanlawson:
Adds a way to test native shadow DOM in `@lwc/jest-preset`. (I guess it's "native" because it's implemented by Jest/JSDOM. 🙂 ). Fixes #93.

## Options file

To add this option, I had to add a config file, `lwc-jest.config.js`. It can contain:

```js
module.exports = {
  nativeShadow: true
}
```

Why `lwc-jest.config.js`? Well, I looked at a few Jest presets mentioned on [the Jest website](https://jestjs.io/). Here's what they're doing:

- [jest-mongodb](https://github.com/shelfio/jest-mongodb): `jest-mongodb-config.js`
- [jest-dynamodb](https://github.com/shelfio/jest-dynamodb): `jest-dynamodb-config.js`
- [jest-puppeteer](https://github.com/smooth-code/jest-puppeteer): `jest-puppeteer.config.js`

I think `lwc-jest.config.js` looks better than `lwc-jest-config.js` (especially next to `jest.config.js`), so I went with `jest-puppeteer`'s format. 🙂 

## Testing

In our tests, I'd like to confirm that we're actually reading the config file and running in the right shadow mode. The most straightforward way I found is to run the full test suite in both synthetic and shadow mode using `process.env.NATIVE_SHADOW` to switch.

This has the benefit of ensuring our tests don't break in either mode. I can also confirm that both modes are working correctly by sniffing `ShadowRoot.prototype.constructor.toString()`.

## Gus WI

W-9318249


pmdartus:
I am a little concerned with creating yet another configuration file for LWC. We already have today the [`lwc.config.json`](https://developer.salesforce.com/blogs/2020/09/lightning-web-components-module-resolution.html) configuration. 

Here are a couple of other alternatives worth considering IMO:
- Using jest `global` options to configure the preset. This approach is similar to [ts-jest](https://kulshekhar.github.io/ts-jest/docs/getting-started/options).
- Exposing multiple presets. Today the preset is exposed via `@lwc/jest-preset`, we can for example also expose `@lwc/jest-preset/synthetic-shadow` and `@lwc/jest-preset/native-shadow`.

nolanlawson:
I'm happy to go with the `ts-jest` approach. It's definitely less machinery to maintain (e.g. do we maintain `lwc-jest.config.cjs` for CommonJS in an ESM project? 😛  ). The multi-package solution seems a bit less clean to me, especially given we may add other options later.

nolanlawson:
OK, switched to using Jest `globals`.

.circleci/config.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       - load_workspace
       - run:
           name: Run unit tests
-          command: yarn test --runInBand
```
pmdartus, line 71: It is fine to run the tests in parallel locally. On the CI we measured that running the tests sequentially [is faster](https://jestjs.io/docs/26.x/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server). This is why the `--runInBand` only applies to the CI.

package.json
```diff
@@ -3,13 +3,14 @@
     "private": true,
     "description": "LWC test plugins and utilities",
     "scripts": {
-        "test": "jest --no-cache",
+        "test": "run-s test:synthetic test:native",
+        "test:synthetic": "jest --no-cache --runInBand",
+        "test:native": "cross-env NATIVE_SHADOW=1 jest --no-cache --runInBand",
```
pmdartus, line 8: Using the jest project feature would allow running the same tests in both synthetic and native shadow without having to run two independent scripts, which I find more elegant:

```js
// jest.config.js
function unitTest({ nativeShadow }) {
    return {
        displayName: {
            name: `unit (${nativeShadow ? 'native' : 'synthetic'} shadow)`,
            color: nativeShadow ? 'blue' : 'cyan'
        },

        rootDir: '<rootDir>/packages',
        testMatch: ['**/__tests__/**/?(*.)(test).js'],

        globals: {
            'lwc-jest': {
                nativeShadow
            }
        },
    };
}

function integration({ nativeShadow }) {
    return {
        displayName: {
            name: `integration (${nativeShadow ? 'native' : 'synthetic'} shadow)`,
            color: nativeShadow ? 'blue' : 'cyan'
        },

        rootDir: '<rootDir>/test',
        preset: '@lwc/jest-preset',
        moduleNameMapper: {
            '^smoke/(.+)$': '<rootDir>/src/modules/smoke/$1/$1',
        },

        globals: {
            'lwc-jest': {
                nativeShadow
            }
        },
    };
}

module.exports = {
    projects: [
        unitTest({ nativeShadow: false }),
        unitTest({ nativeShadow: true }),
        integration({ nativeShadow: false }),
        integration({ nativeShadow: true })
    ],
};
```

packages/@lwc/jest-preset/README.md
```diff
@@ -4,32 +4,54 @@ Tools to assist with testing Lightning Web Components (LWC) with Jest. This proj
...
+ ```
 
-3. Update the `moduleNameMapper` entry in your Jest config to point to where your LWC components live. For example, use the following to map all components in the `example` and `other` namespaces:
+#### nativeShadow
 
-    ```json
-    {
-        "moduleNameMapper": {
-            "^(example|other)/(.+)$": "<rootDir>/src/test/modules/$1/$2/$2"
+Optionally, you can configure `@lwc/jest-preset` to use native shadow DOM rather than synthetic shadow DOM. To do so, add the following to `jest.config.js`:
```
pmdartus, line 39:
```suggestion
By default, this preset is configured to run the tests with synthetic shadow DOM. Optionally, you can configure `@lwc/jest-preset` to use native shadow DOM rather than synthetic shadow DOM. To do so, add the following to `jest.config.js`:
```

packages/@lwc/jest-preset/README.md
```diff
@@ -4,32 +4,54 @@ Tools to assist with testing Lightning Web Components (LWC) with Jest. This proj
 
 ## Usage
 
-### Jest Preset Configuration
+### Installation
 
-1. Install `@lwc/synthetic-shadow`. This is a polyfill for ShadowRoot that was tailor-made for LWC.
+```shell
+yarn add --dev @lwc/jest-preset @lwc/synthetic-shadow
```
pmdartus, line 10: This preset not only has a `peerDependency` on `@lwc/synthetic-shadow` but also `@lwc/compiler` and `@lwc/engine-dom`. 

We either need to update the documentation or update the peer dependency list, but this can be done in another PR.

packages/@lwc/jest-preset/README.md
```diff
@@ -4,32 +4,54 @@ Tools to assist with testing Lightning Web Components (LWC) with Jest. This proj
...
+```
 
-    `yarn add --dev @lwc/synthetic-shadow`
+Optionally, you can omit `@lwc/synthetic-shadow` if you want to test native shadow DOM rather than the polyfill. (See [`nativeShadow`](#nativeshadow) below.)
```
pmdartus, line 13: > Optionally, you can omit `@lwc/synthetic-shadow` if you want to test native shadow DOM rather than the polyfill.

It is not ideal as Yarn warns about missing dependencies, npm@6 errors out during install when a peer dependency is missing and npm@7 automatically install missing peer dependencies.

.circleci/config.yml
nolanlawson: @pmdartus It doesn't surprise me that `--runInBand` would be faster on CI since it probably has fewer cores than our local dev boxes. I'll see if I can get it to do `--runInBand` only on CI.

packages/@lwc/jest-preset/README.md
nolanlawson: You're right, let's remove this. It's an unnecessary optimization.

.circleci/config.yml
nolanlawson: Jest `projects` makes this possible, no prob.

package.json
nolanlawson: Done, thanks. TIL

test/src/modules/smoke/shadow/__tests__/shadow.spec.js
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+describe('shadow DOM', () => {
+    // Smoke test to check that we're running in the right shadow DOM mode
+    if (global['lwc-jest'].nativeShadow) {
+        it('should be using native shadow DOM', () => {
+            // sniff for JSOM's ShadowRoot implementation
+            expect(ShadowRoot.prototype.constructor.toString()).toContain('class ShadowRoot extends globalObject.DocumentFragment');
```
ekashida, line 13: If we need to rely on internal details, might as well only rely on our own internal details?
```suggestion
            expect(ShadowRoot.prototype.constructor.toString()).not.toContain('function SyntheticShadowRoot');
```

pmdartus (approved):

packages/@lwc/jest-preset/README.md
```diff
@@ -4,32 +4,52 @@ Tools to assist with testing Lightning Web Components (LWC) with Jest. This proj
...
+By default, this preset is configured to run the tests with synthetic shadow DOM. Optionally, you can configure `@lwc/jest-preset` to use native shadow DOM rather than synthetic shadow DOM. To do so, add the following to `jest.config.js`:
 
-    ```json
-    {
-        "moduleNameMapper": {
-            "^(example|other)/(.+)$": "<rootDir>/src/test/modules/$1/$2/$2"
+```json
+{
+    "globals": {
+        "lwc-jest": {
```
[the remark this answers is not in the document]
nolanlawson: I went with `lwc-jest` to follow the pattern from `ts-jest`. TBH no option is really great because with `ts-jest` it's the same as the name of the package, so a bit easier to remember, whereas in this case the package name is `@lwc/jest-preset` which would be awkward to put in a global.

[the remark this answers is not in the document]
pmdartus: Jest preset doesn't provide any standard configuration hook, this is why we are adding a new global object. As this configuration flag is only for `jest`, I am fine with `lwc-jest`.

ekashida (approved):

PR #125, opened 2021-06-28
feat(jest-transformer): transform apex methods to a valid wire adapter

jodarove:
This PR modifies the jest-transformer to return a TestApexWireAdapter for apex (and apexContinuation) methods whenever `@salesforce/wire-service-jest-util` [library](https://github.com/salesforce/wire-service-jest-util) is available in the test environment. Also, when `@salesforce/wire-service-jest-util` is present, the apex method will be a `jest.fn()` with a mock implementation returning a resolved promise.

This PR also [fixes a bug](https://github.com/salesforce/lwc-test/pull/125/commits/1bbdd74d94defebe959632431a8d6221fd8deca6) when transforming apex (and apexContinuation) resources:

1. When modules A and B consume the same apex method, but via different import specifiers, they should reference the same apex method.
```js
// A
import apexMethod from '@salesforce/apex/Foo.fooMethod';
// B
import fooMethod from '@salesforce/apex/Foo.fooMethod';

// A(apexMethod) should be the same as B(fooMethod)
```
2. When modules A and B consume different apex methods, but via the same import specifier, they should reference different apex methods. Ex:
```js
// A
import apexMethod from '@salesforce/apex/Foo.fooMethod';
// B
import apexMethod from '@salesforce/apex/Bar.barMethod';

// A(apexMethod) should be different than B(apexMethod)
```

### DIFF
--- a/packages/@lwc/jest-transformer/src/transforms/utils.js
+++ b/packages/@lwc/jest-transformer/src/transforms/utils.js
@@ -69,14 +69,23 @@ function stringScopedImportTransform(t, path, importIdentifier, fallbackData) {
  * id, it must be the same object in the component under test and the test case
  * itself. Due to this requirement, we save the mock to the global object to be
  * shared.
+ *
+ * When @salesforce/wire-service-jest-util is available, we will try to resolve
+ * to a valid wire (apex) adapter definition.
  */
 const resolvedPromiseTemplate = babelTemplate(`
     let RESOURCE_NAME;
     try {
         RESOURCE_NAME = require(IMPORT_SOURCE).default;
-    } catch (e) {
-        global.MOCK_NAME = global.MOCK_NAME || function RESOURCE_NAME() { return Promise.resolve(); };
-        RESOURCE_NAME = global.MOCK_NAME;
+    } catch (importSourceNotDefined) {
+        try {
+            const { createApexTestWireAdapter } = require('@salesforce/wire-service-jest-util');
+            global.MOCK_NAME = global.MOCK_NAME || createApexTestWireAdapter(jest.fn().mockImplementation(() => Promise.resolve()));
+            RESOURCE_NAME = global.MOCK_NAME;
+        } catch (wireServiceJestUtilNotDefined) {
+            global.MOCK_NAME = global.MOCK_NAME || function RESOURCE_NAME() { return Promise.resolve(); };
+            RESOURCE_NAME = global.MOCK_NAME;
+        }
     }
 `);
 
--- a/packages/@lwc/jest-transformer/src/transforms/utils.js
+++ b/packages/@lwc/jest-transformer/src/transforms/utils.js
@@ -78,14 +78,18 @@ const resolvedPromiseTemplate = babelTemplate(`
     try {
         RESOURCE_NAME = require(IMPORT_SOURCE).default;
     } catch (importSourceNotDefined) {
+        global._apexMethodsModuleRegistryHack = global._apexMethodsModuleRegistryHack || {};
+
         try {
             const { createApexTestWireAdapter } = require('@salesforce/wire-service-jest-util');
-            global.MOCK_NAME = global.MOCK_NAME || createApexTestWireAdapter(jest.fn().mockImplementation(() => Promise.resolve()));
-            RESOURCE_NAME = global.MOCK_NAME;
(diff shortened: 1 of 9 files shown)


packages/@lwc/jest-transformer/src/transforms/utils.js
```diff
@@ -69,14 +69,27 @@ function stringScopedImportTransform(t, path, importIdentifier, fallbackData) {
...
  */
 const resolvedPromiseTemplate = babelTemplate(`
     let RESOURCE_NAME;
     try {
         RESOURCE_NAME = require(IMPORT_SOURCE).default;
-    } catch (e) {
-        global.MOCK_NAME = global.MOCK_NAME || function RESOURCE_NAME() { return Promise.resolve(); };
-        RESOURCE_NAME = global.MOCK_NAME;
+    } catch (importSourceNotDefined) {
+        global._apexMethodsModuleRegistryHack = global._apexMethodsModuleRegistryHack || {};
```
pmdartus, line 81: Let's remove the `hack` term from the identifier.

```suggestion
        global._apexMethodsModuleRegistry = global._apexMethodsModuleRegistry || {};
```

test/src/modules/transformer/apex/__tests__/apex.test.js
```diff
@@ -4,12 +4,30 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { ApexMethod, refreshApex, getSObjectValue } from '../apex';
+import { ApexMethod, BarMethod, refreshApex, getSObjectValue } from '../apex';
+import { ApexMethod as NotSameApexMethod } from '../otherApexConsumer';
+import FooMethod from '@salesforce/apex/FooClass.FooMethod';
```
pmdartus, line 9: To make this test easier to understand could your rename the `ApexMethod` export to with the actual method name?

```suggestion
import { FooMethod, BarMethod, refreshApex, getSObjectValue } from '../apex';
import { BarMethod as BarMethodFromOther } from '../otherApexConsumer';
import FooMethodDirect from '@salesforce/apex/FooClass.FooMethod';
```

Or even better using namespace import to make it easier to trace each object to the appropriate module.

```suggestion
import * as apex from '../apex';
import * as otheApex from '../otherApexConsumer';
import FooMethod from '@salesforce/apex/FooClass.FooMethod';
```

test/src/modules/transformer/apex/__tests__/apex-wire-jest-utils.test.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    it('passes a jest.fn() to createApexTestWireAdapter', () => {
+        expect(ApexMethod.mock.calls).toHaveLength(0);
+        ApexMethod();
+        expect(ApexMethod.mock.calls).toHaveLength(1);
+    });
+
+    it('it has resolved promise as default implementation', () => {
+        return ApexMethod().then(() => expect(true).toBe(true));
```
pmdartus, line 39:
```suggestion
        await expect(ApexMethod()).resolves.toBe(undefined);
```

test/src/modules/transformer/apex/__tests__/apex-wire-jest-utils.test.js
```diff
@@ -0,0 +1,42 @@
...
+    it('has different identities for different apex classes', () => {
+        expect(ApexMethod).not.toBe(BarMethod);
+    });
+
+    it('called createApexTestWireAdapter from @salesforce/wire-service-jest-util', () => {
+        expect(ApexMethod.adapter).toBe(true);
+    });
+
+    it('passes a jest.fn() to createApexTestWireAdapter', () => {
+        expect(ApexMethod.mock.calls).toHaveLength(0);
```
pmdartus, line 33: How does it work when the same adapter is used multiple times in a test suite? I don't think the mock will reset between tests which means that it will keep past invocation in the `.mock.calls`.

jodarove:
Closing this PR, was intended as a fix for https://github.com/salesforce/sfdx-lwc-jest/issues/232 but the team decided to keep the existing functionality (revert automock at the resolver level) and leave test authors to implement their own apex methods mocks.

PR #139, opened 2021-11-22
fix: fix typescript components with no HTML

nolanlawson:
There is [an error](https://github.com/salesforce/luvio/pull/781#discussion_r753566072) that appears if you use TypeScript for component files, and if those components have no explicit template HTML files:

```
Cannot find module './foo.html' from 'src/components/foo.ts'
```

This PR fixes the error and adds a test to verify.


wes566 (approved):

pmdartus (approved):
Looks good to me. 

~This code should be refactored once we land the [programmatic style injection RFC](https://github.com/salesforce/lwc-rfcs/blob/pmdartus/css-extensibility/text/0000-programmatic-style-injection.md#lwc-stylesheet-import-enforcement). We will then have a clear signal whenever a stylesheet is optional or not.~

----

**Edit:** Please ignore the previous comment. The programmatic style injection RFC defines how to differentiate between required and optional stylesheets. It doesn't cover optional stylesheets.
 
However, the previous comment made me realize that we can extend the principle defined in [LWC stylesheet import enforcement](https://github.com/salesforce/lwc-rfcs/blob/pmdartus/css-extensibility/text/0000-programmatic-style-injection.md#lwc-stylesheet-import-enforcement) to HTML templates. We are effectively in a situation where we have to guess if a template is optional or not. We could also make the compiler append a `&optional=true` query to all the implicit template imports to indicate that the requested template is optional.

nolanlawson:
OK, let's cover that in future work. We should definitely have one central place where we define all this resolution logic, so that Jest/Rollup/Webpack/etc can share it.

PR #152, opened 2022-05-20
Add support for Jest@28

pmdartus:
This PR adds support for Jest@28. 

One major callout. With version 28, Jest stopped shipping `jest-environment-jsdom` by default ([post](https://jestjs.io/blog/2022/04/25/jest-28#jsdom-19)). Consumers are now required to install this package separately. I don't think it's worth a major version bump, but it's something we should call out in the release note.


nolanlawson (approved):

nolanlawson:
Agreed this is not a breaking change. IMO this is a `feat:`.

PR #156, opened 2022-08-05
feat: lwc/jest-ssr-preset

jodarove:
# Details

This PR adds a new jest preset to `@lwc/jest-preset` package which that allows testing server rendering of an LWC component using jest. Check the `README.md` for usage.

The `@lwc/jest-preset/ssr` also provides a new snapshot serializer to pretty-print SSR output to make it easier to analyze larger HTML files. Compared to the existing `@lwc/jest-preset` it also resolves CSS files (including scoped css) as they are used to generate the SSR output.

### Brief explanation of how it works:

In LWC, the component code is the same for the client and the server, what changes are the `'lwc'` module that is injected into the component (`@lwc/engine-dom`/`@lwc/engine-server`), and the environment (dom/node) in which the code runs.

The jest-preset/ssr simply ensures that when used, `lwc` resolves to `@lwc/engine-server` and the jest environment is `node`. Additionally, the name of `ssr` tests must end in `.ssr-(spec|test).(ts|js)`.

**Note:** Wire adapters are not (currently) supported on the server, therefore no special support was added for [`@salesforce/wire-service-jest-util`](https://github.com/salesforce/wire-service-jest-util).


yarn.lock
```diff
@@ -1633,6 +1633,11 @@
   resolved "https://registry.yarnpkg.com/@lwc/engine-dom/-/engine-dom-2.20.3.tgz#5723491ab73896fc28a17785095ec27d5ffc1acb"
   integrity sha512-G1+e5Vm2TFIJgY5FobKNpp/WdWLd2janUl9bqR453vhzMYexexKo60qFxm/LsRZ2jcXcI7zMtQFmXgy5mCyEiQ==
 
+"@lwc/engine-server@^2.20.3":
+  version "2.21.1"
```
jodarove, line 1637: nts: as is, this PR gives a warning because the template is compiled with an older version. the lwc versions needs to be updated before merging this PR

packages/@lwc/jest-ssr-preset/ssr-resolver.js
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+const lwcResolver = require('@lwc/jest-resolver');
+
+const WHITELISTED_LWC_PACKAGES = {
```
nolanlawson, line 9:
```suggestion
const ALLOWLISTED_LWC_PACKAGES = {
```

packages/@lwc/jest-ssr-preset/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "@lwc/compiler": "*",
+        "@lwc/engine-server": "*",
+        "jest": "^26 || ^27 || ^28"
+    },
+    "dependencies": {
+        "@lwc/jest-resolver": "11.4.0",
+        "@lwc/jest-transformer": "11.4.0"
+    },
+    "engines": {
+        "node": ">=10"
```
nolanlawson, line 33: Do we really support Node 10? I'm a bit surprised. CircleCI executes Node 14, so that's all we can be sure of.

packages/@lwc/jest-ssr-preset/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```json
+{
+    "moduleNameMapper": {
+        "^(example|other)/(.+)$": "<rootDir>/src/test/modules/$1/$2/$2"
+    }
+}
+```
+
+**_1:_** Notice that a jest config only allows one preset per configuration. In order to allow client and server jest tests to live alongside, you might consider creating a new configuration, eg: create `jest-ssr.config.js` and run jest with the `--config` [option](https://jestjs.io/docs/cli#--configpath) to run ssr tests.
```
nolanlawson, line 35: It might be good to have some code examples of this. I'm sure a lot of people are not familiar with configuring Jest this way.

packages/@lwc/jest-ssr-preset/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "keywords": [
+        "jest",
+        "lwc"
+    ],
+    "files": [
+        "/src/*.js"
+    ],
+    "peerDependencies": {
+        "@lwc/compiler": "*",
+        "@lwc/engine-server": "*",
```
nolanlawson, line 25: Can we scope this down a bit more? Something like `2^`?

packages/@lwc/jest-ssr-preset/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "url": "https://github.com/salesforce/lwc-test/issues"
+    },
+    "license": "MIT",
+    "main": "jest-preset.js",
+    "keywords": [
+        "jest",
+        "lwc"
+    ],
+    "files": [
+        "/src/*.js"
```
nolanlawson, line 21: I'm not sure this is right. The JS files are not in `src` so shouldn't this just be `*.js`?

packages/@lwc/jest-ssr-preset/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "directory": "packages/@lwc/jest-ssr-preset"
+    },
+    "bugs": {
+        "url": "https://github.com/salesforce/lwc-test/issues"
+    },
+    "license": "MIT",
+    "main": "jest-preset.js",
+    "keywords": [
+        "jest",
+        "lwc"
```
nolanlawson, line 18:
```suggestion
        "lwc",
        "ssr",
        "server"
```

test/src/modules/ssr/basic/__tests__/basic.ssr-test.js
```diff
@@ -0,0 +1,20 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { renderComponent } from 'lwc';
+import Basic from '../basic';
+
+it('renders a basic component', () => {
+    const renderedComponent = renderComponent('resolver-basic', Basic, { msg: 'Hello world' });
+
+    expect(renderedComponent).toMatchInlineSnapshot(
+        `"<resolver-basic><template shadowroot=\\"open\\"><h1>Basic, Hello world</h1></template></resolver-basic>"`
```
nolanlawson, line 14: Maybe I'm wrong, but I imagine most folks will be using file (not inline) snapshots, and it seems odd to me that this isn't pretty-printed so that it's easy to debug large HTML files. A big HTML snapshot file with no newlines would be difficult to read.

Should we offer some kind of automatic pretty-printing? Or at least recommend a pretty-printer to users in the `README`?

packages/@lwc/jest-ssr-preset/README.md
```diff
@@ -0,0 +1,48 @@
+# @lwc/jest-ssr-preset
```
nolanlawson, line 1: Should it be named `@lwc/jest-preset-ssr`? When I [search for jest presets on npm](https://www.npmjs.com/search?q=jest%20preset), I see a lot of names like `jest-preset-angular`, `jest-preset-gatsby`, `jest-preset-typescript`, etc. Also we have `@lwc/engine-dom` and `@lwc/engine-server`. So even if it's not "correct" English, it seems like the convention to put the modifier at the end.

packages/@lwc/jest-ssr-preset/package.json
jodarove: tbh, I just added the same as the existing `@lwc/jest-preset`, but jest 26 and [lwc](https://github.com/salesforce/lwc/blob/master/package.json#L79) also support node 10. we can always change it on the repo on a later/side PR.

packages/@lwc/jest-ssr-preset/ssr-resolver.js
jodarove: also opened #157 to fix this issue across the repo.

test/src/modules/ssr/basic/__tests__/basic.ssr-test.js
jodarove: This test is internal, since it is small, I prefer to do it inline since IMHO is better for maintainability. is up to devs to use `.toMatchSnapshot`, `.toMatchInlineSnapshot`, etc.

About the automatic pretty-printing, I think is a great suggestion. A new serializer is provided in https://github.com/salesforce/lwc-test/pull/156/commits/83e22d17f72aa92fa1fdd4267a63509edc0b5449 based on https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-server/src/__tests__/fixtures.spec.ts#L54-L117.

The benefit of this functionality implemented as a serializer is that there is no new API that the devs must learn, the drawback is that it will treat every matched text as HTML, but that's probably the point when using `@lwc/jest-preset-ssr`, that's why I think is acceptable.

The functionality can also be used as a new custom matcher, ex: `expect(HTML).toMatchServerSnapshot()`. what is your (and other reviewers) opinion?

nolanlawson: This is awesome!!

nolanlawson (approved):

test/src/modules/ssr/basic/__tests__/basic.ssr-test.js
```diff
@@ -0,0 +1,32 @@
...
+    expect(renderedComponent).toMatchSnapshot();
+});
+
+it('renders a basic component and saves inline formatted snapshot', () => {
+    const renderedComponent = renderComponent('resolver-basic', Basic, { msg: 'Hello world' });
+
+    expect(renderedComponent).toMatchInlineSnapshot(`
+        <resolver-basic>
+          <template shadowroot="open">
+            <h1>
```
jodarove, line 22: iirc, this snapshot should contain a style tag. is that still the case?

nolanlawson: You're right! Something is strange here. There should definitely be a `<style>` here. 🤔

nolanlawson: Ahhh I think I know why. In Jest, we don't actually resolve the CSS files – we just return an empty string.

https://github.com/salesforce/lwc-test/blob/d8893f9105459b9adc91cec7e50a7207ded7d156/packages/%40lwc/jest-resolver/src/index.js#L50-L53

We should probably fix this for the SSR preset to work properly.

nolanlawson: Scoped styles will also need their own special resolution to work correctly.

jodarove: ah, jajaja, you are too fast! thanks! that [should](https://github.com/salesforce/lwc-test/pull/156#discussion_r939181595) be it. ill let you know once i verify.

will also check about scoped styles.

jodarove: @nolanlawson https://github.com/salesforce/lwc-test/pull/156/commits/799d0ec8cfcec09b801ead6bc9ffe66a198663ba provides support for styles and scoped styles for the ssr preset. thanks for pointing the direction!

packages/@lwc/jest-preset-ssr/README.md
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+``` 
+
+
+### Testing
+
+Create a `__tests__` inside the bundle of the LWC component under test.
+
+Then, create a new test file in `__tests__` that follows the naming convention `<js-file-under-test>.ssr-test.js`. See an example in this projects `src/test/ssr` directory.
```
ravijayaramappa, line 53:
```suggestion
Then, create a new test file in `__tests__` that follows the naming convention `<js-file-under-test>.ssr-test.js`. See an example in this project's `src/test/ssr` directory.
```

packages/@lwc/jest-preset-ssr/package.json
```diff
@@ -0,0 +1,37 @@
...
+        "ssr",
+        "server"
+    ],
+    "files": [
+        "/src/*.js"
+    ],
+    "peerDependencies": {
+        "@lwc/compiler": "^2",
+        "@lwc/engine-server": "^2",
+        "jest": "^26 || ^27 || ^28"
```
ravijayaramappa, line 28: We [discussed](https://salesforce-internal.slack.com/archives/G0210EW0V2R/p1658425992729429) plans to drop support for jest26 for `@lwc/jest-preset`. Since this is a new package, jest26 should be skipped for it.
```suggestion
        "jest": ^27 || ^28"
```

packages/@lwc/jest-preset-ssr/jest-preset.js
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+module.exports = {
+    // Jest 26 default testEnvironment is jsdom: https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults
```
ravijayaramappa, line 8: Related to Jest26 comment below
```suggestion
```

packages/@lwc/jest-preset-ssr/src/ssr-html-serializer.js
```diff
@@ -0,0 +1,104 @@
...
+            const isVoid = isVoidElement(tagNameMatch[0], HTML_NAMESPACE);
+            const isClosing = src.charAt(pos + 1) === '/';
+            const isComment =
+                src.charAt(pos + 1) === '!' &&
+                src.charAt(pos + 2) === '-' &&
+                src.charAt(pos + 3) === '-';
+
+            start = pos;
+            while (src.charAt(pos++) !== '>') {
+                // Keep advancing until consuming the closing tag.
```
ravijayaramappa, line 69:
```suggestion
                // Keep advancing until the closing tag.
```

packages/@lwc/jest-preset-ssr/src/ssr-resolver.js
```diff
@@ -0,0 +1,52 @@
...
+    if (isScoped) {
+        fileName = basename(fileName, '.scoped');
+    }
+    const absPath = resolve(basedir, importee);
+    const dir = dirname(absPath);
+    const jsFile = join(dir, fileName + '.js');
+    const tsFile = join(dir, fileName + '.ts');
+
+    return (
+        // if it is an css file
```
ravijayaramappa, line 29:
```suggestion
        // if it is a css file
```

packages/@lwc/jest-preset-ssr/src/ssr-resolver.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+module.exports = function (path, options) {
+    const effectivePath = path.endsWith('?scoped=true') ? path = path.substring(0, path.length - 12) : path;
+
+    if (ALLOWLISTED_LWC_PACKAGES[effectivePath]) {
+        return options.defaultResolver(require.resolve(ALLOWLISTED_LWC_PACKAGES[effectivePath]), options);
+    }
+
+    // If it is a CSS and it exist, we should parse it since it must be rendered in ssr. The default resolver will
```
ravijayaramappa, line 45:
```suggestion
    // If it is a CSS and it exists, we should parse it since it must be rendered in ssr. The default resolver will
```

ravijayaramappa (approved):
Minor nits

jest.config.js
```diff
@@ -48,6 +48,18 @@ module.exports = {
         unitTest({ nativeShadow: false }),
         unitTest({ nativeShadow: true }),
         integration({ nativeShadow: false }),
-        integration({ nativeShadow: true })
+        integration({ nativeShadow: true }),
```
nolanlawson, line 51: Just to check: this doesn't break if we have `nativeShadow: false`, right? I imagine it shouldn't affect things for SSR, but you never know...

packages/@lwc/jest-preset-ssr/package.json
nolanlawson: Should we do that in a separate PR? It seems orthogonal to this PR, and would requite updating the Circle CI tests as well.

packages/@lwc/jest-transformer/src/index.js
```diff
@@ -84,6 +84,15 @@ function transformTypeScript(src, filePath) {
     return code;
 }
 
+function getScopedStylesOption(src, filePath) {
+    const ext = path.extname(filePath);
+    const isCSS = ext === '.css';
+    let fileName = path.basename(filePath, '.css');
+    const isScoped = path.extname(fileName) === '.scoped';
```
nolanlawson, line 91: Technically we're "supposed" to use `?scoped=true` as the signal that CSS is scoped. Although this will probably change due to [programmatic style injection](https://github.com/salesforce/lwc-rfcs/pull/56#discussion_r736579693) and we already do it "incorrectly" in `lwc-webpack-plugin`, so it probably doesn't matter.

nolanlawson (approved):
Small nits, LGTM

packages/@lwc/jest-preset-ssr/package.json
ravijayaramappa: Sounds good.

packages/@lwc/jest-preset-ssr/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```json
+{
+    "moduleNameMapper": {
+        "^(example|other)/(.+)$": "<rootDir>/src/test/modules/$1/$2/$2"
+    }
+}
+```
+
+**_1:_** Notice that a jest config only allows one preset per configuration. In order to allow client and server jest tests to live alongside, you might consider creating a new configuration.
```
pmdartus, line 35: It took me quite some time to understand what `**_1:_**` was doing here.

IMO, this is one of the key takeaways of the `README`. I would recommend creating a dedicated section speaking about this, eg: **LWC DOM and SSR component test setup**

packages/@lwc/jest-preset-ssr/README.md
```diff
@@ -0,0 +1,59 @@
+# @lwc/jest-preset-ssr
+
+Tools to assist with testing Lightning Web Components (LWC) rendered on the server with Jest. This project provides 2 services: preset Jest configuration for testing Lightning web components, and stubs for common external libraries used in Lightning web components.
+
+## Usage
+
+### Installation
+
+```shell
+yarn add --dev @lwc/jest-preset-ssr @lwc/compiler @lwc/engine-server
```
pmdartus, line 10: I am not a big fan of asking to install a separate package seperate package for SSR testing. This is yet another package consumers will have to upgrade and keep in sync in their `package.json`.

What do you think about exposing the SSR config via `@lwc/jest-preset/ssr`?

packages/@lwc/jest-transformer/src/index.js
jodarove: so, I wasn't able to make the default resolver load correctly the scoped CSS file with the `?scoped=true`. I would prefer to leave this as is and open an issue so we can tackle it with time later on, if we really want to push for it.

packages/@lwc/jest-preset-ssr/README.md
jodarove: Great suggestion! makes sense. Initially, I wanted to only use `@lwc/jest-preset`, but because jest only allows one jest preset, it would not be possible; I didn't think about this option.

packages/@lwc/jest-preset/README.md
```diff
@@ -52,10 +59,70 @@ By default, this preset is configured to run the tests with synthetic shadow DOM
 }
 ```
 
+#### LWC components rendered on the Server
+
+Add the `@lwc/jest-preset/ssr` preset to the Jest configuration of your `package.json` (also see [LWC DOM and SSR component test setup](#LWC-DOM-and-SSR-component-test-setup) section) like so:
```
pmdartus, line 64:
```suggestion
Add the `@lwc/jest-preset/ssr` preset to the Jest configuration like so:
```

packages/@lwc/jest-preset/README.md
```diff
@@ -52,10 +59,70 @@ By default, this preset is configured to run the tests with synthetic shadow DOM
...
+```json
+{
+  "jest": {
+    "preset": "@lwc/jest-preset/ssr",
+    "moduleNameMapper": {
+      "^(example|other)/(.+)$": "<rootDir>/src/test/modules/$1/$2/$2"
+    }
+  }
+}
+```
```
pmdartus, line 85: I don't think it is worth mentioning this again.

```suggestion
```

pmdartus (approved):

PR #158, opened 2022-08-30
fix: add support for jest 29

nolanlawson:
Adds supports for Jest 29 in a non-breaking way. Also adds tests in CI to confirm that Jest 29 tests are passing.


test/src/modules/serializer/dom/__tests__/input.spec.js
```diff
@@ -1,19 +0,0 @@
...
-    ['null', null],
-    ['empty object', {}],
-    ['without the nodeType property', { anotherKindOfProperty: 5 }],
-    ['with an element node type', document.createElement('div')],
-    ['with a text node type', document.createTextNode('')],
-    ['with a comment node type', document.createComment('')],
-    ['with another kind of node type', document.createAttribute('foo')],
-])('should return the expected state when we have %s', (useCase, input) => {
-    expect(input).toMatchSnapshot();
-});
```
nolanlawson, line 19: This test is not possible to maintain using snapshots that work in both Jest 28 and Jest 29, because the Jest snapshot format changed. I'm also not sure of the value of this test, since it seems to just be testing the snapshot serialization for a variety of JS and JSDOM types.

nolanlawson: And yes, I also tried to figure out how we could have snapshots that worked in both Jest 28 and 29, and I was unable.

jye-sf: I'm wondering if that was the original intent of this test: to catch situations like this when the jest snapshot format changes so that we're alerted to the breaking change.

jye-sf: So, I'm also not sure of the value of this test, but it does run through both of jest-serializer's main [exports](https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-serializer/src/index.js#L106). For thoroughness, I'm concerned about removing it outright unless we know all the use cases are covered by other existing tests.

Is the entire thing failing, or are there cases that work in both Jest 28 and 29? Can we, at minimum, add those back in?

jye-sf (changes requested):

test/src/modules/serializer/dom/__tests__/input.spec.js
nolanlawson: The way Jest snapshots work is that, even if we just have them for one Jest version (e.g. behind an `if`), then Jest will complain about outdated snapshots for the other variants.

That said, our other variants only run in CI, and I think it's just a warning. So yeah, lemme see if I can add it just for Jest 29.

jye-sf: I dug into it a bit more, and these tests seem to be unit tests for the `test()` function in index.js. They end up cycling through that function (used by Jest) testing all possible input types for `obj` (which makes sense given the name of the test). Most of these inputs are also not covered by any of the other tests in serializer, and this function is used directly by Jest so we should definitely keep test coverage.

Unfortunately, the main part of that function resolves around objects, which is also the breaking change in Jest 29.

If we write these as unit tests, that would bypass the issues with Jest version upgrades. What are your thoughts on that?

jye-sf: If anyone's curious, here's the documentation around these functions and how Jest uses them: https://github.com/facebook/jest/blob/main/packages/pretty-format/README.md#writing-plugins

jye-sf (approved):
Thanks for putting them back in!

These would probably be better as unit tests, but I don't have a preference whether that's on this PR or a separate PR.

test/src/modules/serializer/dom/__tests__/input.spec.js
nolanlawson: I switched to inline snapshots. It looks like we can use these and have snapshot tests that run in all Jest versions.

nolanlawson:
@jye-sf Switched to snapshot tests, so now the original test is preserved. :)

jye-sf (approved):

PR #160, opened 2022-09-26
fix: render CSS correctly on both client and server

nolanlawson:
Currently, we only render CSS in server mode, not client mode. We should really render the CSS in both cases.

This moves the CSS rendering logic from `@lwc/jest-preset/src/ssr/resolver.js` to `@lwc/jest-resolver/src/index.js` and makes both the client and server resolver rely on the same logic.

### DIFF
--- a/packages/@lwc/jest-resolver/src/index.js
+++ b/packages/@lwc/jest-resolver/src/index.js
@@ -37,6 +37,29 @@ function isImplicitHTMLImport(importee, { basedir }) {
     );
 }
 
+function isValidCSSImport(importee, { basedir }) {
+    const ext = extname(importee);
+    const isCSS = ext === '.css';
+    let fileName = basename(importee, '.css');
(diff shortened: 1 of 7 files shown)


packages/@lwc/jest-resolver/src/index.js
```diff
@@ -37,6 +37,29 @@ function isImplicitHTMLImport(importee, { basedir }) {
     );
 }
 
+function isValidCSSImport(importee, { basedir }) {
+    const ext = extname(importee);
+    const isCSS = ext === '.css';
+    let fileName = basename(importee, '.css');
+    const isScoped = extname(fileName) === '.scoped';
```
nolanlawson, line 44: Technically we should be relying on the presence of `?scoped=true` rather than `.scoped.css` to determine whether something is scoped or not. But I couldn't figure out how to do that in Jest, so I'm punting on that in this PR.

PR #162, opened 2022-10-13
fix: allow @imports in CSS to use moduleNameMapper

nolanlawson:
Fixes https://github.com/salesforce/sfdx-lwc-jest/issues/288.

This allows you to do `@import 'foo/bar'` in your CSS and have it resolve to a file somewhere called `foo/bar.css`'. Unfortunately the way to get this to work is to use `moduleNameMapper`, but we already have that documented here:

https://github.com/salesforce/lwc-test/blob/5a9f6f0e72c12b64324f08476c77f62e57ee08a7/packages/%40lwc/jest-preset/README.md?plain=1#L37-L41


packages/@lwc/jest-resolver/src/index.js
```diff
@@ -40,26 +40,21 @@ function isImplicitHTMLImport(importee, { basedir }) {
...
         isCSS &&
         // the css file must exist
-        fs.existsSync(absPath) &&
-        // there must be a js/ts file with the same name in the same folder
-        (fs.existsSync(jsFile) || fs.existsSync(tsFile))
+        fs.existsSync(absPath)
     );
 }
 
+function isValidImport(importee, { basedir }) {
```
jye-sf, line 53: nit: to distinguish from `isValidCSSImport`?
```suggestion
function isValidScriptImport(importee, { basedir }) {
```

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -70,14 +65,30 @@ function getLwcPath(path, options) {
...
+    // If the extension is empty, try to infer it
+    if (extname(path) === '') {
+        if (isValidImport(path + '.ts', options) || isValidImport(path + '.js', options)) {
+            return path // the resolution algo will automatically add '.ts'/'.js' as necessary
+        }
+
+        // If there is no extension, try to infer a .css path. We do a special check for CSS to handle @imports inside of
+        // CSS files, which can be something like `"foo/bar"` resolving to `"foo/bar.css"`, but unlike '.js', Node's
+        // resolution algorithm doesn't automatically add the '.css'
+        // TODO: this will fail if there is a .js file in the same directory as the .css file. Need a way to distinguish
```
jye-sf, line 87: If there's a `.js` file in the same directory, we would have returned the JS file at line 81 above right?

Is there potentially a way to determine what file we're importing from, and is it a safe assumption to say if there's no extension, we default to the same type of file as the importer?

jye-sf (approved):

packages/@lwc/jest-resolver/src/index.js
nolanlawson: I think we would have to switch based on the importee. Because you would be able to import that JS file if you were doing so from another JS file.

That said, I don't actually know if the platform/OSS resolver supports this, and therefore if this is actually a concern in practice. So far the only example I've seen was of an orphaned `.css` file, no corresponding `.js` file.

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -70,14 +65,30 @@ function getLwcPath(path, options) {
...
+        // between `@import "foo/bar"` in CSS and `import "foo/bar"` in JS
```
nolanlawson, line 88:
```suggestion
        // between `@import "foo/bar"` in CSS and `import "foo/bar"` in JS (assuming platform 
        // compiler supports this)
```

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -70,14 +65,31 @@ function getLwcPath(path, options) {
...
     if (extname(path) === '.css' && !isValidCSSImport(path, options)) {
         return EMPTY_CSS_MOCK;
     }
 
-    // If is an implicit imported html (auto-binded from the compiler) return an empty file
-    if (isImplicitHTMLImport(path, options)) {
-        return EMPTY_HTML_MOCK;
+    // If the extension is empty, try to infer it
+    if (extname(path) === '') {
+        if (isValidImport(path + '.ts', options) || isValidImport(path + '.js', options)) {
```
nolanlawson, line 80:
```suggestion
        if (isValidScriptImport(path + '.ts', options) || isValidScriptImport(path + '.js', options)) {
```

packages/@lwc/jest-resolver/src/index.js
nolanlawson: Modified the comment to note that I don't know if we actually need to solve this.

PR #170, opened 2023-01-06
fix: use ?scoped=true as signal for scoped CSS

nolanlawson:
Currently `lwc-test` is the only case where we use `.scoped.css` in the filename as a signal for whether a CSS file is scoped or not. Technically we should be using the presence of `?scoped=true`.

In Jest, I could not figure out a way to plumb the query param through from the resolver to the transformer. Research on how to do this turned up nothing.

The simplest solution I could find is to create a new package, `@lwc/jest-shared`, which serves as a communication mechanism between `@lwc/jest-resolver` and `@lwc/jest-transformer`. One notes the presence of `?scoped=true`, whereas the other looks up a filename later to see if it should be treated as scoped.

### DIFF
--- a/packages/@lwc/jest-resolver/src/index.js
+++ b/packages/@lwc/jest-resolver/src/index.js
@@ -6,6 +6,8 @@
  */
 const fs = require('fs');
 const { resolve, extname, join, dirname, basename } = require('path');
+const { URLSearchParams } = require('url')
+const { addKnownScopedCssFile } = require('@lwc/jest-shared');
 
 const EMPTY_CSS_MOCK = resolve(__dirname, '..', 'resources', 'emptyStyleMock.js');
 const EMPTY_HTML_MOCK = resolve(__dirname, '..', 'resources', 'emptyHtmlMock.js');
@@ -55,10 +57,17 @@ function isValidScriptImport(importee, { basedir }) {
     return fs.existsSync(absPath)
 }
 
+function parseForQueryParams(path) {
+    const [ filename, search ] = path.split('?')
+    const params = new URLSearchParams(search);
+    return { filename, params };
+}
+
 function getLwcPath(path, options) {
-    if (path.endsWith('?scoped=true')) {
-        // remove query param for scoped styles
-        path = path.substring(0, path.length - 12);
+    const { filename, params } = parseForQueryParams(path);
+    path = filename; // remove query param for scoped styles (`?scoped=true`)
+    if (params.get('scoped') === 'true') {
+        addKnownScopedCssFile(resolve(options.basedir, path));
     }
     // If is a special LWC package, resolve it from commonjs
     if (ALLOWLISTED_LWC_PACKAGES[path]) {
--- a/packages/@lwc/jest-resolver/src/index.js
+++ b/packages/@lwc/jest-resolver/src/index.js
@@ -6,7 +6,7 @@
  */
 const fs = require('fs');
 const { resolve, extname, join, dirname, basename } = require('path');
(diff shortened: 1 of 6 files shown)


packages/@lwc/jest-resolver/src/index.js
```diff
@@ -55,10 +57,17 @@ function isValidScriptImport(importee, { basedir }) {
...
+    const params = new URLSearchParams(search);
+    return { filename, params };
+}
+
 function getLwcPath(path, options) {
-    if (path.endsWith('?scoped=true')) {
-        // remove query param for scoped styles
-        path = path.substring(0, path.length - 12);
+    const { filename, params } = parseForQueryParams(path);
+    path = filename; // remove query param for scoped styles (`?scoped=true`)
```
nolanlawson, line 68: This is where we would run into problems. If we _don't_ strip out the query param here, then Jest will throw an error later because it can't find the file.

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -55,10 +57,17 @@ function isValidScriptImport(importee, { basedir }) {
     return fs.existsSync(absPath)
 }
 
+function parseForQueryParams(path) {
+    const [ filename, search ] = path.split('?')
```
nolanlawson, line 61: There is a chance that the filename contains a `?` character, but I could not actually write a test for this case because Jest itself threw an error. So I'm going to assume that Jest doesn't support `?` characters in filenames and we don't need to worry.

Even if there is a filename in a parent/grandparent folder, we don't care, because this function is only parsing the immediate path, which is taken directly from the `import` statement, e.g. `'./stylesheet.css'`.

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -6,6 +6,8 @@
  */
 const fs = require('fs');
 const { resolve, extname, join, dirname, basename } = require('path');
+const { URLSearchParams } = require('url')
```
rui-rayqiu, line 9: minor: missing semicolon

packages/@lwc/jest-resolver/src/index.js
```diff
@@ -55,10 +57,20 @@ function isValidScriptImport(importee, { basedir }) {
     return fs.existsSync(absPath)
 }
 
+function parseForQueryParams(path) {
+    // There is a chance that the filename contains a ? character, but Jest itself throws an error in this case.
+    // Even if there is a ? in a parent/grandparent folder, this function only parses the immediate path (e.g.
+    // `./filename.css`), so it doesn't matter.
+    const [ filename, search ] = path.split('?')
```
rui-rayqiu, line 64: minor: missing semicolon

rui-rayqiu (approved):
LGTM!

packages/@lwc/jest-resolver/src/index.js
nolanlawson:
```suggestion
const { URLSearchParams } = require('url');
```

packages/@lwc/jest-resolver/src/index.js
nolanlawson:
```suggestion
    const [ filename, search ] = path.split('?');
```

nolanlawson:
I would like a quick +1 from @pmdartus or @ravijayaramappa to confirm they are cool with adding a new package here. (`@lwc/jest-shared`)

pmdartus (approved):
I spent some time on the Jest documentation, but I haven't found another alternative. LGTM.

PR #177, opened 2023-03-28
feat(jest-transformer): enable new dynamic components

jmsjtu:
This PR enables [dynamic components](https://github.com/salesforce/lwc-rfcs/blob/master/text/0133-dynamic-component-creation.md) for testing.

This PR only enables dynamic components, a follow-up PR will be needed to retrieve the namespace and name to properly create a dynamic component with the correct tag name which is a breaking change.

In 244, the users of dynamic components are internal developers and a pilot group of ISVs.

### DIFF
--- a/test/src/modules/smoke/dynamicCmp/__tests__/__snapshots__/dynamicCmp.test.js.snap
+++ b/test/src/modules/smoke/dynamicCmp/__tests__/__snapshots__/dynamicCmp.test.js.snap
@@ -1,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`should render dynamically imported component 1`] = `
+<smoke-dynamic-import>
+  #shadow-root(open)
+    <x-test>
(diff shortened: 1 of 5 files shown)


test/src/modules/smoke/dynamicCmp/__tests__/__snapshots__/dynamicCmp.test.js.snap
```diff
@@ -0,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`should render dynamically imported component 1`] = `
+<smoke-dynamic-import>
+  #shadow-root(open)
+    <x-test>
```
jmsjtu, line 6: For now all dynamic components will have the `x-test` tag name, this should be ok though as component authors can use an attribute such as `data-*` or `class` to select the dynamic component.

ravijayaramappa (approved):

PR #178, opened 2023-03-30
chore: update lwc to v2.41.3 and other dependencies

ravijayaramappa:


nolanlawson:
Looks like Nexus snuck into the lockfile.

nolanlawson (approved):

PR #184, opened 2023-06-05
fix(serializer): replace LWC scope tokens

nolanlawson:
This is a PoC of replacing the scope tokens in serialized templates. In serialized HTML, all scope tokens are replaced with `__lwc_scope_token__`.

This should eventually allow us to implement lowercasing of scope tokens (https://github.com/salesforce/lwc/issues/3313) because it would no longer be a breaking change, since the snapshot will not change if the scope token changes.

### DIFF
--- a/packages/@lwc/jest-transformer/src/index.js
+++ b/packages/@lwc/jest-transformer/src/index.js
@@ -7,8 +7,9 @@
 const path = require('path');
 const crypto = require('crypto');
 
-const { isKnownScopedCssFile } = require('@lwc/jest-shared');
+const { isKnownScopedCssFile, addKnownScopeToken } = require('@lwc/jest-shared');
 
+const MagicString = require('magic-string');
 const babelCore = require('@babel/core');
 const lwcCompiler = require('@lwc/compiler');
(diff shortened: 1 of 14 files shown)


packages/@lwc/jest-transformer/src/index.js
```diff
@@ -112,7 +113,39 @@ module.exports = {
         // **Note: .html and .css don't return valid sourcemaps cause they are used for rollup
         const config = map && map.version ? { inputSourceMap: map } : {};
 
-        return babelCore.transform(code, { ...BABEL_CONFIG, ...config, filename });
+        let result = babelCore.transform(code, { ...BABEL_CONFIG, ...config, filename });
+
+        let matcher;
+        const scopeToken = filePath.endsWith('.html') && (matcher = code.match(/tmpl.stylesheetToken = "([^"]+)";/)) && matcher[1];
```
nolanlawson, line 119: Once https://github.com/salesforce/lwc/pull/3556 is merged, we can get rid of this regex code.

nolanlawson:
GitHub is messed up, so I'm opening another PR

PR #185, opened 2023-06-07
fix(serializer): replace LWC scope tokens (BREAKING CHANGE)

nolanlawson:
New version of #184

 In serialized HTML, all scope tokens are replaced with `__lwc_scope_token__`. This is a breaking change because it breaks all existing snapshots, and because we now require the LWC `peerDependency` of 2.48+.


packages/@lwc/jest-preset/package.json
```diff
@@ -22,10 +22,10 @@
...
     ],
     "peerDependencies": {
-        "@lwc/compiler": "*",
-        "@lwc/engine-dom": "*",
-        "@lwc/engine-server": ">=2",
-        "@lwc/synthetic-shadow": "*",
+        "@lwc/compiler": "^2.48.0",
+        "@lwc/engine-dom": "^2.48.0",
+        "@lwc/engine-server": "^2.48.0",
+        "@lwc/synthetic-shadow": "^2.48.0",
```
nolanlawson, line 28: `@lwc/compiler` v2.48.0 add the `cssScopeTokens` return value, which we need. We might as well bump every other LWC peer dep as well, since folks shouldn't be mixing-and-matching different versions of LWC packages.

ekashida (approved):
Looks good--just one question about whether you considered using the existing format of `[lwc:scope-token]`?

nolanlawson:
I like it! Let's do `lwc:scope-token`

nolanlawson:
Ah, unfortunately I just realized we can't do `lwc:scope-token` because the colon messes up the classes in the serialized CSS. This is invalid CSS.

Screenshot 2023-06-07 at 1 23 09 PM

nolanlawson:
On second thought, how about `lwc-xxxxxx`? Since the new format we're going to use is `lwc-<hash>`, maybe it's a little less distracting of a name.

nolanlawson:
Sounds good, let's stick with the first one then!

PR #190, opened 2023-06-22
fix(serializer): remove "-host" from serialized scope token

nolanlawson:
Another thing missed in #187. We need to sort the regex input so that the full `foo-host` string is replaced.

### DIFF
--- a/packages/@lwc/jest-shared/src/index.js
+++ b/packages/@lwc/jest-shared/src/index.js
@@ -46,7 +46,7 @@ function getKnownScopeTokensRegex() {
     // sort from longest to shortest so that `{foo-host}` is fully replaced, not just `{foo}-host`
     const regexString = [...knownScopeTokens]
         .sort((a, b) => b.length - a.length)
-        .join('|')
+        .join('|');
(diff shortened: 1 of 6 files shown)


packages/@lwc/jest-shared/src/index.js
```diff
@@ -43,8 +43,12 @@ function isKnownScopeToken(str) {
  @returns {RegExp} - regex representing the list of known scope tokens
  */
 function getKnownScopeTokensRegex() {
+    // sort from longest to shortest so that `{foo-host}` is fully replaced, not just `{foo}-host`
+    const regexString = [...knownScopeTokens]
+        .sort((a, b) => b.length - a.length)
+        .join('|')
```
divmain, line 49: Semicolon? Sorry to be a pedant.

divmain (approved):

packages/@lwc/jest-shared/src/index.js
nolanlawson: We need to add eslint/prettier. Fixed! 😛

PR #191, opened 2023-06-27
feat: retrieve namespace and name in jest transformer

jmsjtu:
This PR retrieves the namespace and name to use in the jest transformer.

The namespace/name is retrieved through the resolved file path in a similar way to how it is retrieved in the LWC rollup plugin.

If the namespace/name does not resolve to a valid custom element name, the default namespace/name will be `x/test`.

This approach was used to parse the namespace/name for the following reasons:

1. Jest does not provide a way to retrieve the _actual_ import string (moduleNameMapper) used to import a module.

- Retrieving the _actual_ import string will require reverse engineering the mapping logic to infer what it is.

3. Although not ideal, this allows component authors to customize how they want their tag names to appear.
4. Component authors who wish to select the component can use an attribute such as `class` or `data-*`.

There are main drawbacks to this approach is that tag names can appear strangely if the file path doesn't fit nicely

For example, if the file path is `/my/path/to/mocks/__mocks__/`

The namespace is `mocks` and the name is `__mocks__`.


test/src/modules/smoke/dynamicCmp/__tests__/__snapshots__/dynamicCmp.test.js.snap
```diff
@@ -1,13 +1,40 @@
...
     <x-test>
+      #shadow-root(open)
+    </x-test>
+</smoke-dynamic-import>
+`;
+
+exports[`Dynamic components mocking dynamic component should render with non virtual mock 1`] = `
+<smoke-dynamic-import>
+  #shadow-root(open)
+    <dynamic-cmp-__tests__>
```
jmsjtu, line 15: This is an example of how the tag name can appear strangely.

test/src/modules/smoke/dynamicCmp/__tests__/__snapshots__/dynamicCmp.test.js.snap
```diff
@@ -1,13 +1,40 @@
...
+      #shadow-root(open)
+    </dynamic-cmp-__tests__>
+</smoke-dynamic-import>
+`;
+
+exports[`Dynamic components should render dynamically imported component 1`] = `
+<smoke-dynamic-import>
+  #shadow-root(open)
+    <smoke-child>
```
jmsjtu, line 33: This is an example of how the tag name can appear nicely.

jmsjtu:
We discussed internally and decided to continue using the x and test for the namespace + name.

PR #196, opened 2023-07-13
fix: configure `enableLwcSpread` compiler option conditionally

seckardt:
To prevent tons of warning log lines, the LWC compiler option `enableLwcSpread` is now only configured for compiler versions lower than 2.49.1.

The PR that lead to these log lines is https://github.com/salesforce/lwc/pull/3544 ([v2.49.1](https://github.com/salesforce/lwc/releases/tag/v2.49.1)).

### DIFF
--- a/packages/@lwc/jest-transformer/package.json
+++ b/packages/@lwc/jest-transformer/package.json
@@ -30,7 +30,8 @@
         "@babel/preset-typescript": "^7.22.5",
         "@lwc/jest-shared": "12.0.3",
         "babel-preset-jest": "^29.5.0",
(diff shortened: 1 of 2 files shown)


abdulsattar (approved):

nolanlawson:
Thanks for the PR. As an alternative, what if LWC OSS only logged the warning once? This would fix the issue for `lwc-test` as well as other downstreams (e.g. LWR).

nolanlawson (approved):

nolanlawson:
This PR still seems reasonable either way, let's do it.

PR #199, opened 2023-07-28
fix: make snapshots resilient to whitespace changes (BREAKING CHANGE)

### LINKED ISSUE #198: Axe a11y validation error: `(duplicate-id-aria) IDs used in ARIA and labels must be unique`
Our snapshots serialize every ID and `aria-*` idref as `[shadow-guid]`:

https://github.com/salesforce/lwc-test/blob/35157d2926c57b287a3bfa77c895733e0a03a9fa/packages/%40lwc/jest-serializer/src/clean-element-attrs.js#L8

This leads to HTML that [Axe](https://www.deque.com/axe/) complains about because of apparent duplicate IDs:

```html
<input type="text" aria-labelledby="[shadow:guid]">
<label id="[shadow:guid]">label1</label>
<input type="text" aria-labelledby="[shadow:guid]">
<label id="[shadow:guid]">label2</label>
```

We should:

1. At the very least, disable this serialization in native shadow
2. Consider another solution so that Axe doesn't complain

nolanlawson:
Breaking change, don't merge until we have the other breaking change in (which would address #198).

This makes snapshots resilient to empty text nodes, which fixes the extra whitespace caused by https://github.com/salesforce/lwc/pull/3308.

W-13831241

### DIFF
--- a/packages/@lwc/jest-serializer/src/clean-element-text-content.js
+++ b/packages/@lwc/jest-serializer/src/clean-element-text-content.js
@@ -1,0 +1,12 @@
+// LWC VFragments use empty text nodes as start/end markers:
+// https://github.com/salesforce/lwc/blob/50c1bfa/packages/%40lwc/engine-core/src/framework/api.ts#L96-L97
+// Removing these from the serialized snapshot reduces whitespace noise in snapshots.
+function cleanElementTextContent(node) {
+    for (const child of [...node.childNodes]) {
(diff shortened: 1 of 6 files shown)


ekashida:
I think this PR was unintentionally associated as a fixing change to #198

nolanlawson:
You're right, fixed

ekashida (approved):

nolanlawson:
Waiting to merge until the other breaking change is ready

PR #201, opened 2023-08-01
chore: remove id/idref serialization (BREAKING CHANGE)

### LINKED ISSUE #198: Axe a11y validation error: `(duplicate-id-aria) IDs used in ARIA and labels must be unique`

ekashida:
Fixes #198.

Remove id/idref serialization in jest snapshots due to conflicts with a11y testing (i.e., duplicated id values `[shadow:guid]`). LWC's scoped id implementation is no longer likely to change as `@lwc/synthetic-shadow` is in maintenance mode so there's not much value in keeping this around.


rui-rayqiu (approved):

nolanlawson (approved):
LGTM. I was going to suggest we add a test for the new behavior, but I guess it's not very useful?

PR #207, opened 2023-08-15
fix: minimal ARIA string reflection polyfill for JSDOM

nolanlawson:
It's clear to me from Jest tests running in core that a lot of folks took a dependency on LWC's ARIA polyfill, at least for basic string reflection such as `role`/`ariaLabel`/`ariaExpanded`/etc. Unfortunately it's not supported in JSDOM yet (https://github.com/jsdom/jsdom/issues/3323), which makes it hard to drop the polyfill from LWC OSS (https://github.com/salesforce/lwc/pull/3666).

I propose that we put a minimal ARIA polyfill into `@lwc/jest-preset`, implementing _only_ the standard supported in Chrome/Safari. Reason being:

- JSDOM might not merge the fix (https://github.com/jsdom/jsdom/pull/3586) anytime soon
- Even if they do, the latest version of `jest-environment-jsdom` is using [jsdom v20](https://unpkg.com/browse/jest-environment-jsdom@29.6.2/package.json), which is 2 major versions behind (latest is [v22](https://github.com/jsdom/jsdom/releases/tag/22.1.0))
- It would be a lot of effort to fix people's tests in core, and the only impact it would even have outside Jest is to fix Firefox in non-LEX environments

This allows us to at least get the global polyfill out of the core LWC OSS repo, without needlessly breaking a bunch of Jest tests.


test/src/modules/smoke/aria/__tests__/aria.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    'ariaValueMax',
+    'ariaValueMin',
+    'ariaValueNow',
+    'ariaValueText',
+];
+
+describe('ARIA string reflection', () => {
+    for (const prop of ARIA_STRING_PROPS) {
+        // e.g. `ariaPosInSet` -> `aria-posinset`
+        const attribute = prop.replace(/[A-Z]/, (letter) => `-${letter}`).toLowerCase();
```
abdulsattar, line 53: Nit: this line is copied from the code it's testing. If we change ARIA_STRING_PROPS to an object with prop -> attribute (which can be copied directly from https://w3c.github.io/aria/#accessibilityroleandproperties-correspondence), we would be account for exceptions (which there are none as of today, but just in case).

abdulsattar (approved):
LGTM!

test/src/modules/smoke/aria/__tests__/aria.spec.js
nolanlawson: That's a good point; that would be a better test!

nolanlawson: Fixed: 21530ef

PR #210, opened 2023-08-17
feat: add `expect().toThrowInConnectedCallback` matcher

nolanlawson:
After running the Jest tests in core, I realized that a lot of component authors are doing something like this:

```js
expect(() => {
  document.body.appendChild(createComponent()) // throws in connectedCallback
}).toThrow('foo')
```

This doesn't work with native custom element lifecycle, because the error is no longer thrown locally, as described  in https://github.com/salesforce/lwc/pull/3662.

I believe component authors need a simple way to assert on errors thrown in `connectedCallback`s. So I propose adding a new matcher:

```js
expect(() => {
  ...
}).toThrowInConnectedCallback('foo')
```

This PR adds this new matcher. This matcher works both with LWC components and with native custom elements. It doesn't matter whether the LWC components are running in native lifecycle mode or synthetic lifecycle mode.

This should make it a lot easier for component authors to upgrade their tests. I tested on a half-dozen core Jest tests, and using `toThrowInConnectedCallback`, I was able to fix a lot of test failures.

### DIFF
--- a/packages/@lwc/jest-preset/README.md
+++ b/packages/@lwc/jest-preset/README.md
@@ -109,3 +109,43 @@ Create a `__tests__` inside the bundle of the LWC component under test.
 Then, create a new test file in `__tests__` that follows the naming convention `<js-file-under-test>.test.js` for DOM tests and `<js-file-under-test>.ssr-test.js` for ssr tests. See an example in this projects `src/test` directory.
 
 Now you can write and run the Jest tests!
+
+### Custom matchers
+
+This package contains convenience functions to help test web components, including Lightning Web Components.
+
+#### expect().toThrowInConnectedCallback
+
+Allows you to test for an error thrown by the `connectedCallback` of a web component. `connectedCallback` [does not necessarily throw errors synchronously](https://github.com/salesforce/lwc/pull/3662), so this utility makes it easier to test for `connectedCallback` errors.
+
+```js
+// Component
+export default class Throws extends LightningElement {
+    connectedCallback() {
+        throw new Error('whee!');
+    }
+}
+```
+
+```js
+// Test
+import { createElement } from 'lwc';
+
+it('Should throw in connectedCallback', () => {
+    const element = createElement('x-throws', { is: Throws });
+    expect(() => {
+        document.body.appendChild(element);
+    }).toThrowErrorInConnectedCallback(/whee!/);
+});
+```
+
+The argument passed in to `toThrowInConnectedCallback` behaves the same as for [Jest's built-in `toThrow`](https://jestjs.io/docs/expect#tothrowerror):
+
+-   Regular expression: error message matches the pattern.
+-   String: error message includes the substring.
+-   Error object: error message is equal to the message property of the object.
+-   Error class: error object is instance of class.
+
+#### expect().toThrowErrorInConnectedCallback
+
+Equivalent to `toThrowInConnectedCallback`.
--- a/packages/@lwc/jest-preset/README.md
+++ b/packages/@lwc/jest-preset/README.md
@@ -124,6 +124,8 @@ import '@lwc/jest-preset';
 
 Allows you to test for an error thrown by the `connectedCallback` of a web component. `connectedCallback` [does not necessarily throw errors synchronously](https://github.com/salesforce/lwc/pull/3662), so this utility makes it easier to test for `connectedCallback` errors.
 
+##### Example
+
 ```js
 // Component
 export default class Throws extends LightningElement {
@@ -145,6 +147,8 @@ it('Should throw in connectedCallback', () => {
 });
 ```
 
+##### Error matching
+
 The argument passed in to `toThrowInConnectedCallback` behaves the same as for [Jest's built-in `toThrow`](https://jestjs.io/docs/expect#tothrowerror):
 
 -   Regular expression: error message matches the pattern.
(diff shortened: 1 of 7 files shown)


nolanlawson:
I looked around to see if anyone else had implemented something similar for other web component libraries (Lit, Stencil, etc.), but didn't find anything in a quick search.

packages/@lwc/jest-preset/src/matchers/expect-throw-in-connected-callback.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+    if (typeof matcher === 'function') {
+        // Error class
+        return error instanceof matcher;
+    }
+    if (matcher.message) {
+        // Error object
+        return error.message === matcher.message;
+    }
+}
```
nolanlawson, line 24: I really really didn't want to write this junk myself, and would prefer to just delegate to Jest's `toThrow` matcher, but apparently that's not possible: https://github.com/jestjs/jest/issues/2547#issuecomment-348417525

nolanlawson:
Got TypeScript types working: c7ab61c

Screenshot 2023-08-18 at 8 38 07 AM

jmsjtu (approved):
This LGTM, I added a small comment about including additional info in the `README`. 

I also though about any potential risks by introducing this API surface but I couldn't think of any compelling reasons not to do it.

jmsjtu (approved):
This LGTM, I added a small comment about including additional info in the `README`. 

I also though about any potential risks by introducing this API surface but I couldn't think of any compelling reasons not to do it.

packages/@lwc/jest-preset/README.md
```diff
@@ -109,3 +109,52 @@ Create a `__tests__` inside the bundle of the LWC component under test.
...
+Note that, for these matchers to work properly in TypeScript, you must import this package from your `*.spec.ts` files:
+
+```js
+import '@lwc/jest-preset';
+```
+
+#### expect().toThrowInConnectedCallback
+
+Allows you to test for an error thrown by the `connectedCallback` of a web component. `connectedCallback` [does not necessarily throw errors synchronously](https://github.com/salesforce/lwc/pull/3662), so this utility makes it easier to test for `connectedCallback` errors.
+
```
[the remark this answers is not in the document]
nolanlawson: Done!

nolanlawson:
Thanks for the review @jmsjtu! I would like one more +1, just since this is adding new API surface so it's kinda risky. @ekashida?

packages/@lwc/jest-preset/types/index.d.ts
```diff
@@ -0,0 +1,48 @@
...
+     * @description
+     * Assert that appending a custom element to the DOM causes an error to be thrown from that element's `connectedCallback`.
+     * @example
+     * expect(() => {
+     *     document.body.appendChild(element);
+     * }).toThrowInConnectedCallback('expected error message');
+     * @see
+     * [@lwc/jest-preset docs](https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-preset/README.md#custom-matchers)
+     */
+    toThrowInConnectedCallback(expected?: unknown): R;
```
ravijayaramappa, line 19: Why not use the same types as Jest's `toThrow`
```suggestion
    toThrowInConnectedCallback(expected?: string | Constructable | RegExp | Error): R;
```

packages/@lwc/jest-preset/types/index.d.ts
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * Equivalent to `toThrowInConnectedCallback`.
+     * @example
+     * expect(() => {
+     *     document.body.appendChild(element);
+     * }).toThrowErrorInConnectedCallback('expected error message');
+     * @see
+     * [@lwc/jest-preset docs](https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-preset/README.md#custom-matchers)
+     */
+    toThrowErrorInConnectedCallback(expected?: unknown): R;
```
ravijayaramappa, line 33:
```suggestion
    toThrowErrorInConnectedCallback(expected?: string | Constructable | RegExp | Error): R;
```

packages/@lwc/jest-preset/types/index.d.ts
nolanlawson: `unknown` is [actually what they use in Jest](https://github.com/jestjs/jest/blob/fb7d95c8af6e0d65a8b65348433d8a0ea0725b5b/packages/expect/src/types.ts#L345-L352).

I'm not sure why they do that though TBH. I'm fine with changing it to be stricter.

nolanlawson: Also what is `Constructable`? Is it [from `vitest`](https://github.com/vitest-dev/vitest/blob/c03faa2204dc758900d5dbab732af83ee947e4e1/packages/expect/src/types.ts#L135C1-L136)?

nolanlawson: [Found it in `vitest`](https://github.com/vitest-dev/vitest/blob/dff60e82804ea2eb17aeff6752c1fa53b5f910b8/packages/utils/src/types.ts#L23-L25)

I think overall though I'd rather align with Jest, since otherwise people may run into issues when they change `toThrow()` to `toThrowInConnectedCallback()`.

nolanlawson: Nevermind, I'm totally wrong; [it's in `@types/jest`](https://unpkg.com/browse/@types/jest@29.5.3/index.d.ts). I'll fix this.

packages/@lwc/jest-preset/types/index.d.ts
nolanlawson: Fixed in 4023598

packages/@lwc/jest-preset/types/index.d.ts
nolanlawson: Fixed in e0e70bb

ravijayaramappa (approved):

PR #212, opened 2023-09-20
chore: transform `import.meta.env.SSR` to `false` within component code

divmain:
The `import.meta.env.SSR` flag is used in component code where the component author needs to vary behavior depending on the execution environment. In production, `import.meta.env.SSR` will be either `true` (if the component is being server-rendered) or `false` (if the component is rendering in the browser).

We do not intend for Jest to be used to test SSR & component hydration. For that reason, `import.meta.env.SSR` should always evaluate to `false` in Jest tests. This PR makes that change.

### DIFF
--- a/packages/@lwc/jest-transformer/src/index.js
+++ b/packages/@lwc/jest-transformer/src/index.js
@@ -94,6 +94,10 @@ module.exports = {
             src = transformTypeScript(src, filePath);
         }
 
+        // Replace `import.meta.env.SSR` with `false` prior to parsing the
+        // JavaScript inside the LWC compiler.
+        src = src.replace(/\bimport\.meta\.env\.SSR\b/g, 'false');
+
         // Set default module name and namespace value for the namespace because it can't be properly guessed from the path
         const { code, map, cssScopeTokens } = lwcCompiler.transformSync(src, filePath, {
             name: 'test',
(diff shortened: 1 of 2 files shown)


packages/@lwc/jest-transformer/src/index.js
```diff
@@ -94,6 +94,10 @@ module.exports = {
             src = transformTypeScript(src, filePath);
         }
 
+        // Replace `import.meta.env.SSR` with `false` prior to parsing the
+        // JavaScript inside the LWC compiler.
+        src = src.replace(/\bimport\.meta\.env\.SSR\b/g, 'false');
+
```
nolanlawson, line 100: Is it worth adding a test for this? We have both SSR and client-side tests in this repo.

nolanlawson:
> We do not intend for Jest to be used to test SSR & component hydration.

We [do have SSR testing capabilities](https://github.com/salesforce/lwc-test/tree/master/packages/%40lwc/jest-preset#lwc-components-rendered-on-the-server).

seckardt:
@divmain Just saw that we were both adding solutions for the very same issue ;) I can only say that the assumption of your PR isn't really true. We already have repositories that are set up to both test SSR and CSR behavior via Jest. That's why we should make sure that this is taken into account.

PR #216, opened 2023-11-08
chore: update serializer to remove empty comments introduced in lwc v5

divmain:
This PR changes the behavior of the serializer to strip `<!---->` comments from snapshots. These comment nodes were introduced as part of salesforce/lwc#3846 and release in `lwc@5.0.0`. We don't want to unnecessarily break everybody's snapshot tests, which is why we're introducing the changes in this PR.

There wasn't an existing test that broke with the upgrade to lwc@5, other than the SSR tests (which will just have to break, as the comment nodes are meaningful when serialized to actual HTML markup). I've added a test here that breaks if you remove the behavioral changes.


ravijayaramappa (approved):

jye-sf (approved):

nolanlawson:
@divmain 

> the comment nodes are meaningful when serialized to actual HTML markup

This may be the case, but in the past we have removed such framework-specific markup from both the SSR and non-SSR snapshots. This removes the amount of boilerplate that a component author has to understand when reading a snapshot file. For example, we previously removed the ZWJ characters from the SSR snapshots.

Do you think it would make sense to remove the empty comments from both the SSR snapshots and the non-SSR snapshots?

PR #217, opened 2023-11-09
feat: add support for `import.meta` properties @W-14387292

seckardt:
This PR adds support for `import.meta` properties inside components that are unit tested via Jest. To do so `import.meta.*` properties are rewritten to `process.*`. Special treatment is added for boolean `import.meta.env.*` properties due to the fact that `process.env.*` only allows/supports string properties.

### DIFF
--- a/packages/@lwc/jest-transformer/src/transforms/import-meta.js
+++ b/packages/@lwc/jest-transformer/src/transforms/import-meta.js
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+module.exports = function ({ types: t }) {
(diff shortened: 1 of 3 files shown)


packages/@lwc/jest-transformer/src/transforms/import-meta.js
```diff
@@ -0,0 +1,42 @@
...
+                    t.isIdentifier(parent.property) &&
+                    parent.property.name === 'env'
+                ) {
+                    // `import.meta.env.*` properties require some special treatment. Unfortunately
+                    // `process.env` only supports `string` properties, which means that `boolean`
+                    // environmental information e.g. `import.meta.env.SSR` do not work as expected
+                    // out-of-the-box without some extra "massaging".
+                    const envPath = Array.from(currentMemberExpressions.values()).at(-2);
+                    const envName = envPath?.get?.('property.name')?.node;
+                    envPath?.replaceWithSourceString?.(`/true/i.test(process.env['${envName}'])`);
```
divmain, line 35: This is a clever workaround!

divmain (approved):

PR #236, opened 2024-04-04
fix(serializer): normalize class attributes

nolanlawson:
Split into two commits to make the changes more obvious.

TODO: do the same for `style` attrs.

W-15365974


jmsjtu (approved):

jmsjtu:
See #237 for `style` attribute changes.

PR #237, opened 2024-04-04
fix(serializer): normalize style attributes

jmsjtu:
Split into two commits to make the changes more obvious.

See #236 for class changes.

BREAKING CHANGE: snapshots may change.

W-15365974

### DIFF
--- a/packages/@lwc/jest-serializer/src/clean-element-attrs.js
+++ b/packages/@lwc/jest-serializer/src/clean-element-attrs.js
@@ -23,6 +23,18 @@ function cleanElementAttributes(elm) {
             elm.removeAttribute(name);
             elm.setAttribute('__lwc_scope_token__', value);
         }
+
+        if (name === 'style') {
+            // Normalize trivial whitespace differences (e.g. `style="  foo   bar "` becomes `style="foo bar"`
+            // and `style=""` is just removed).
+            const styleAttribute = elm.getAttribute(name) || '';
+            const normalizedStyleAttribute = styleAttribute.replace(/\s+/g, ' ').trim();
+            if (normalizedStyleAttribute) {
+                elm.setAttribute(name, normalizedStyleAttribute);
+            } else {
+                elm.removeAttribute(name);
+            }
+        }
     }
 }
 
--- a/packages/@lwc/jest-serializer/src/clean-element-attrs.js
+++ b/packages/@lwc/jest-serializer/src/clean-element-attrs.js
@@ -23,18 +23,6 @@ function cleanElementAttributes(elm) {
             elm.removeAttribute(name);
             elm.setAttribute('__lwc_scope_token__', value);
         }
-
-        if (name === 'style') {
-            // Normalize trivial whitespace differences (e.g. `style="  foo   bar "` becomes `style="foo bar"`
-            // and `style=""` is just removed).
-            const styleAttribute = elm.getAttribute(name) || '';
-            const normalizedStyleAttribute = styleAttribute.replace(/\s+/g, ' ').trim();
-            if (normalizedStyleAttribute) {
-                elm.setAttribute(name, normalizedStyleAttribute);
-            } else {
-                elm.removeAttribute(name);
-            }
-        }
     }
 }
 
--- a/packages/@lwc/jest-serializer/src/clean-element-styles.js
+++ b/packages/@lwc/jest-serializer/src/clean-element-styles.js
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+function cleanElementStyles(elm) {
+    // Normalize trivial whitespace differences (e.g. `style="  foo   bar "` becomes `style="foo bar"`
+    // and `style=""` is just removed).
+    const styleAttribute = elm.getAttribute('style') || '';
+    const normalizedStyleAttribute = styleAttribute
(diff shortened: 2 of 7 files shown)


test/src/modules/serializer/styleAttr/__tests__/__snapshots__/styleAttr.spec.js.snap
```diff
@@ -0,0 +1,89 @@
...
+      style="color:blue;text-align:center"
+    />
+    <div
+      style="color:blue; text-align:center"
+    />
+    <div
+      style="color: blue; text-align: center"
+    />
+    <div />
+    <div />
```
nolanlawson, line 23: Let's add tests for `!important` and extra ending `;`s as well, and whitespace around each.

packages/@lwc/jest-serializer/src/clean-element-attrs.js
```diff
@@ -23,6 +23,18 @@ function cleanElementAttributes(elm) {
...
+            // Normalize trivial whitespace differences (e.g. `style="  foo   bar "` becomes `style="foo bar"`
+            // and `style=""` is just removed).
+            const styleAttribute = elm.getAttribute(name) || '';
+            const normalizedStyleAttribute = styleAttribute.replace(/\s+/g, ' ').trim();
+            if (normalizedStyleAttribute) {
+                elm.setAttribute(name, normalizedStyleAttribute);
+            } else {
+                elm.removeAttribute(name);
+            }
+        }
```
nolanlawson, line 37: We should probably split on `;` and normalize whitespace for each, since there can be arbitrary whitespace before/after the semicolon.

nolanlawson: Also this would be cleaner to go in its own `cleanElementStyle` function/file imo.

nolanlawson: Also: this is not going to handle zany things like `content` or `grid-template-areas` which may contain spaces or semicolons in the value, but I'm not sure that that's worth handling. It quickly devolves into prettier/full CSS parsing, and we probably just need a lightweight formatter.

test/src/modules/serializer/styleAttr/__tests__/__snapshots__/styleAttr.spec.js.snap
jmsjtu: Added in the latest commit! 6d6e6ef557f0da3f8ecb578bc3921642caa4ec33

packages/@lwc/jest-serializer/src/clean-element-styles.js
```diff
@@ -0,0 +1,28 @@
...
+    const styleAttribute = elm.getAttribute('style') || '';
+    const normalizedStyleAttribute = styleAttribute
+        // Remove leading/trailing space
+        .trim()
+        // Replace multiple space, tab, line break with single space
+        .replace(/[\s\n\t]+/g, ' ')
+        .split(';')
+        // Preserve space character following a ';', style="color: blue; text-align: center"
+        .map((attr) => attr.trimEnd())
+        .join(';');
```
nolanlawson, line 20:
```suggestion
        .split(';')
        .map(_ => _.trim())
        .filter(Boolean)
        .map(_ => `${_};`)
        .join(' ');
```

This avoids the duplicates `;`s at the end.

wjhsf:
```suggestion
        .split(';')
        .map(_ => `${_.trim()};`)
        .filter(_ => _ !== ';')
        .join(' ');
```

Same but one less loop.

jmsjtu: This will also append a ';' at the end of a style:

```html
<div style="  color: blue  "></div>
```

Becomes:

```html
<div style="color: blue;"></div>
```

Empty ';' will be removed as well

```html
<div style=";;"></div>
```
 Becomes:

```html
<div />
```

Just want to double-check, are we trying to normalize the whitespace or the style attribute output?

Asking because it won't exactly line up with what the component author will see in the browser.

nolanlawson: Yes, I think we want consistent output. So consistent `;`s at the end, consistent spacing, and just `<div />` if the `style` is semantically empty.

test/src/modules/serializer/styleAttr/__tests__/__snapshots__/styleAttr.spec.js.snap
```diff
@@ -0,0 +1,232 @@
...
+      style="color: blue; text-align: center;"
+    />
+    <div
+      style="color: blue; text-align: center;"
+    />
+    <div
+      style="color: blue; text-align: center;"
+    />
+    <div
+      style="color : blue; text-align : center;"
```
nolanlawson, line 33: This one is going to lead to inconsistencies between static/dynamic sadly... Not sure how to solve without a full CSS parse 🤦

nolanlawson: On second thought, we can't use a CSS parser because we've never validated this content, so it could be syntactically invalid and consumers could be relying on that.

I think we should split on `:` as well and normalize that so that it's always in the format:

```
color: red;
```

or

```
color: red !important;
```

nolanlawson: #238

nolanlawson (approved):

nolanlawson (changes requested):
See above and #239

nolanlawson (approved):

PR #239, opened 2024-04-05
fix: use jsdom to parse style attr and serialize it

nolanlawson:
It occurred to me that, instead of regexes, we can use the `elm.style` CSSStyleDeclaration object to serialize out the `style` attribute. This has some benefits:

1. We delegate the hard work to JSDOM
2. We can handle stuff like `data:` base64 URLs correctly

And some downsides:

1. JSDOM's parsing does not seem perfect, e.g. it doesn't seem to understand `! important` (due to the space) and it doesn't seem to like a leading `;`
2. For invalid CSS declarations (e.g. `color: yolo`) it will just strip them out
3. JSDOM normalizes some values, e.g. `background-color:#FFFFFF;"` becomes `background-color: rgb(255, 255, 255);`

I think this is an acceptable tradeoff, especially since it makes the code a lot cleaner.

### DIFF
--- a/test/src/modules/serializer/styleAttr/__tests__/__snapshots__/styleAttr.spec.js.snap
+++ b/test/src/modules/serializer/styleAttr/__tests__/__snapshots__/styleAttr.spec.js.snap
@@ -164,10 +164,14 @@ exports[`serializes component with different whitespace 1`] = `
     <div
       style="background-color: red !important;"
     />
-    <div />
-    <div />
     <div
-      style="text-align: line-break;"
+      style="color: yellow;"
+    />
+    <div
+      style="color: purple;"
+    />
+    <div
+      style="color: purple; text-align: right;"
     />
     <div />
     <div />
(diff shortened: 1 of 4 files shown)


test/src/modules/serializer/styleAttr/__tests__/__snapshots__/styleAttr.spec.js.snap
```diff
@@ -89,27 +91,34 @@ exports[`serializes component with different whitespace 1`] = `
       style="color: blue; text-align: center;"
     />
     <div
-      style="background-color: red;"
+      style="background-color: red !important;"
```
nolanlawson, line 94: It turns out we don't parse `!important` correctly in some cases. Opened a separate bug for that: https://github.com/salesforce/lwc/issues/4125

jye-sf (approved):
Is the change in behavior for `! important` worth calling out in a comment?

nolanlawson:
`! important` was always kind of broken; we just discovered it now due to actually writing tests for all whitespace combinations.

When looking at snapshots in core, I don't see anyone actually using `! important` so it probably doesn't matter.

PR #260, opened 2024-06-04
Update @salesforce/apex/* imports to return jest.fn()

nwcm:
After looking for ways to  simplify mocking apex methods in a SFDX project, I found this transformer is applied. However the apex method transforms don't make the imports available within jests. 

This change allows for jests import and use apex methods cleanly without having to mock in each test file. 

This doesn't apply to wired apex methods which would need `createApexTestWireAdapter()`  from [@salesforce/sf](https://github.com/salesforce/sfdx-lwc-jest)

```js
// foo.test.js
import apexMethod from "@salesforce/apex/ApexClass.apexMethod";

describe("c-foo", () => {
  afterEach(() => {
    // resetEnvironment();
  });
  
  
  it("Test mock", async () => {
     apexMethod.mockResolvedValue('value');
  }
}
```

Otherwise any apex method requires

```js 
jest.mock(
  "@salesforce/apex/ApexClass.apexMethod",
  () => {
    return {
      default: jest.fn()
    };
  },
  { virtual: true }
);
```

### DIFF
--- a/packages/@lwc/jest-preset/README.md
+++ b/packages/@lwc/jest-preset/README.md
@@ -110,6 +110,36 @@ Then, create a new test file in `__tests__` that follows the naming convention `
 
 Now you can write and run the Jest tests!
 
+#### @Salesforce/apex/ Method Mocks
+
+Imports of `@Salesforce/apex/*` automatically resolve to `jest.fn()`, these can be optionally overwritten.
+
+```js
+import apexMethod from '@Salesforce/apex/apexClass.apexMethod';
+
+it('test apex cal', async () => {
+    apexMethod.mockResolvedValue({ foo: 'bar' });
+});
+```
+
+Optional set function for method manually
+
+```js
+import apexMethod from '@Salesforce/apex/apexClass.apexMethod';
+
+jest.mock(
+    '@salesforce/apex/apexClass.apexMethod',
+    () => ({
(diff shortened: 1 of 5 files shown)


nolanlawson:
Thanks for the contribution! Could you clarify: is this a breaking change?

nolanlawson:
FYI @pozil @Templarian do you have strong opinions about this? Should we just return a `jest.fn()` so that people can mock directly?

pozil:
I like the idea of having a mock directly but I have not clue about the potential impact of such a change.

nolanlawson (changes requested):
I don't think this should be a breaking change. It looks reasonable enough to me.

OTOH could we please add documentation for this so that it's more discoverable? A good place would be [this README](https://github.com/salesforce/lwc-test/blob/master/packages/%40lwc/jest-preset/README.md). Thank you! 🙇

nolanlawson (approved):

nolanlawson (changes requested):
One small change to account for `injectGlobals=false` and I think this is good to go! https://github.com/nwcm/lwc-test/pull/1

nolanlawson (approved):
Thanks a bunch! Really appreciate the time you took to add docs and such. 🙇 Will release this in a bit.

PR #274, opened 2024-07-01
test: ignore expected console error and improve test

nolanlawson:
This removes an annoying error message in the test (which is totally expected) and also improves the tests for apex methods returning a promise.

### DIFF
--- a/test/src/modules/transformer/apex/__tests__/apex.test.js
+++ b/test/src/modules/transformer/apex/__tests__/apex.test.js
@@ -14,8 +14,7 @@ describe('@salesforce/apex/<class>', () => {
 
 describe('@salesforce/apex', () => {
     it('exports refreshApex method returning a promise', () => {
-        expect(refreshApex()).not.toBeUndefined();
-        expect(refreshApex().then).toEqual(expect.any(Function)); // should be a thenable (promise)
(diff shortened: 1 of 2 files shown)


test/src/modules/transformer/apex/__tests__/apex.test.js
```diff
@@ -14,7 +14,9 @@ describe('@salesforce/apex/<class>', () => {
 
 describe('@salesforce/apex', () => {
     it('exports refreshApex method returning a promise', () => {
-        expect(refreshApex()).resolves.toEqual(undefined);
+        expect(refreshApex()).not.toBeUndefined();
+        expect(refreshApex().then).toEqual(expect.any(Function)); // should be a thenable (promise)
```
wjhsf, line 18: Why a thenable rather than instance of Promise?

wjhsf (approved):

test/src/modules/transformer/apex/__tests__/apex.test.js
nolanlawson: Old habit? I guess my mind is still stuck in the days when Promises weren't supported everywhere.

nolanlawson: Fixed: e083fa4

PR #278, opened 2024-07-08
chore(ssr): strip framework-owned attributes

nolanlawson:
Related: https://github.com/salesforce/lwc/pull/4358

We don't want to actually render `data-lwc-host-mutated` in SSR snapshots. This is an internal framework implementation detail that should be stripped.

~~I also had to downgrade Prettier to v2 because apparently Jest inline snapshots don't work with v3. I'm not sure why this didn't fail in https://github.com/salesforce/lwc-test/pull/227 ; maybe because we weren't generating any new snapshots.~~ _Dunno why Prettier v3 wasn't working temporarily; it's working again!_

### DIFF
--- a/packages/@lwc/jest-preset/src/ssr/html-serializer.js
+++ b/packages/@lwc/jest-preset/src/ssr/html-serializer.js
@@ -100,7 +100,14 @@ function formatHTML(src) {
         }
     }
 
-    return res.trim().replace(getKnownScopeTokensRegex(), '__lwc_scope_token__');
+    return (
+        res
+            .trim()
+            .replace(getKnownScopeTokensRegex(), '__lwc_scope_token__')
+            // These special attributes are reserved by the framework and are meaningless to component authors
+            .replace(/ data-lwc-host-mutated/g, '')
+            .replace(/ data-rendered-by-lwc/g, '')
+    );
 }
 
 module.exports = {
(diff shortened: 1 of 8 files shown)


packages/@lwc/jest-preset/src/ssr/html-serializer.js
```diff
@@ -100,7 +100,14 @@ function formatHTML(src) {
...
     }
 
-    return res.trim().replace(getKnownScopeTokensRegex(), '__lwc_scope_token__');
+    return (
+        res
+            .trim()
+            .replace(getKnownScopeTokensRegex(), '__lwc_scope_token__')
+            // These special attributes are reserved by the framework and are meaningless to component authors
+            .replace(/ data-lwc-host-mutated/g, '')
+            .replace(/ data-rendered-by-lwc/g, '')
```
nolanlawson, line 109: We're not currently rendering this in SSR, but we are for CSR.

We probably should render it in SSR, because (sadly) this has effectively become public API surface now, since we weren't thinking about who might take a dependency on this attribute being there.

wjhsf:
> I also had to downgrade Prettier to v2 because apparently Jest inline snapshots don't work with v3.

What doesn't work about v3? I tried running it locally and nothing seemed obviously broken.

nolanlawson:
@wjhsf Try `yarn test -u`

nolanlawson:
Weird, now I can't repro. 🤷

nolanlawson:
OK, un-downgraded Prettier ^

wjhsf (approved):

PR #292, opened 2024-08-21
Jest snapshot-based approach for SSR testing

lesya7:
This PR addresses the following:
1. Create a custom serializer to override the snapshot with the generated SSR markup
2. Create a custom test env for JSDOM env where we introduce the global var that holds an absolute path to a test suite
3. Snapshot utils for SSR/CSR tests
4. Corresponding changes in presets
5. Corresponding changes in jest.config
6. Corresponding changes in the existing tests
7. Unit test coverage for newly introduced APIs


packages/@lwc/jest-jsdom-custom-test-env/package.json
```diff
@@ -0,0 +1,25 @@
...
+        "url": "https://github.com/salesforce/lwc-test.git",
+        "directory": "packages/@lwc/jest-jsdom-custom-test-env"
+    },
+    "bugs": {
+        "url": "https://github.com/salesforce/lwc-test/issues"
+    },
+    "license": "MIT",
+    "main": "src/jsdom-custom-test-env.js",
+    "dependencies": {
+        "jest-environment-jsdom": "^29.7.0"
```
nolanlawson, line 17: This should be a `peerDependency` since its version is tied to the overall Jest version, and we don't know what version that will be since we support multiple.

packages/@lwc/jest-jsdom-custom-test-env/src/__tests__/jsdom-custom-test-env.test.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @jest-environment @lwc/jest-jsdom-custom-test-env
+ */
+
+const path = require('path');
+
+describe('JSDOM custom test environment', () => {
+    it('should correctly set the absolute path of the test file in global.testFilePath', () => {
+        const expectedPath = path.resolve(__filename);
+        expect(global.testFilePath).toBe(expectedPath);
```
nolanlawson, line 10: Do we need the custom environment if the global `__filename` does the same thing?

It's a little more awkward because we would have to pass the filename from the test file into `readSnapshotMarkup`, but I think it might be simpler because we can avoid the custom environment.

packages/@lwc/jest-preset/src/ssr/html-serializer.js
nolanlawson, line 1: The goal of this file is two things:

1. Pretty-print the HTML so that the snapshots are easier to read
2. Remove some framework-specific attributes that are not interesting to the component author (e.g. `data-lwc-host-mutated`, CSS scope tokens).

The value of #<span></span>2 is debatable, but #<span></span>1 is quite nice since it makes snapshots much easier to debug. I can see a case for disregarding the whole thing, though, since it doesn't really do anything you can't already do with an off-the-shelf prettifier like [jest-serializer-html](https://github.com/algolia/jest-serializer-html).

Either way, changing this behavior for all current SSR users would definitely be a breaking change. (We consider it a breaking change when snapshots change.) Is there a way that the new serializer could be used only for SSR+hydration tests?

packages/@lwc/jest-preset/src/ssr/snapshot-serializer.js
```diff
@@ -0,0 +1,36 @@
...
+    },
+
+    /**
+     * Determines if this serializer can handle the given value.
+     *
+     * @param {*} _value - The value to test.
+     * @returns {boolean} A boolean indicating if the serializer can handle the value.
+     */
+    test: function (_value) {
+        return typeof _value === 'string' && _value.length > 0;
```
nolanlawson, line 22: Names like `_value` (with the underscore) are typically used to indicate an unused variable (e.g. to avoid TypeScript lint errors).

packages/@lwc/jest-ssr-snaphot-utils/src/ssr-snapshot-utils.js
```diff
@@ -0,0 +1,113 @@
...
+ * Renders the component's markup, captures it in a snapshot that has a unique snapshot hash.
+ *
+ * @param {string} tagName - The name of the HTML tag or component.
+ * @param {Function} Ctor - The constructor of the Lightning Web Component.
+ * @param {Object} [props={}] - An object representing the properties of the component.
+ * @param {Object} [customTestEnv={}] - An object representing the custom test env where the component is being validated.
+ * @returns {{markup: string, snapshotHash: string}} - An object containing the rendered markup and the generated snapshot hash.
+ */
+function generateAndSnapshotMarkup(tagName, Ctor, props = {}, customTestEnv = {}) {
+    expect.addSnapshotSerializer(jestCustomSnapshotSerializer);
```
nolanlawson, line 17: Can this be done with configuration instead, i.e. [`snapshotSerializers`](https://jestjs.io/docs/configuration#snapshotserializers-arraystring)? It seems odd to call `addSnapshotSerializer` over and over.

test/package.json
```diff
@@ -1,8 +1,10 @@
 {
     "name": "test",
     "version": "16.1.0",
-    "private": "true",
```
nolanlawson, line 4: This is not intended to be a published package.

test/package.json
```diff
@@ -1,8 +1,10 @@
...
     "scripts": {
         "test": "jest"
+    },
+    "dependencies": {
+        "@lwc/ssr-snapshot-utils": "16.1.0"
```
nolanlawson, line 8: I don't think this is strictly required since this is a monorepo.

packages/@lwc/jest-jsdom-custom-test-env/src/__tests__/jsdom-custom-test-env.test.js
lesya7: It's just a unit test for my change that makes sure we can get an absolute path to a test suite during execution. In this test I directly access it but it should, also, be accessible from test utils called from a test.

packages/@lwc/jest-jsdom-custom-test-env/package.json
lesya7: Sure, I'll add "jest-environment-jsdom": "^28 || ^29"

packages/@lwc/jest-preset/src/ssr/html-serializer.js
lesya7: It will be utilized with @lwr/preset-ssr and while calling generateAndSnapshotMarkup. Either way it's going to be a breaking change. I'll look into [jest-serializer-html](https://github.com/algolia/jest-serializer-html).

packages/@lwc/jest-ssr-snaphot-utils/src/ssr-snapshot-utils.js
lesya7: In my case it's not quite feasible because I need to use jestCustomSnapshotSerializer to call an API and pass in SSR-generated markup to replace the original snapshot content. If I don't have to override the snapshot content, then tweaking a config is enough.

test/package.json
lesya7: Oh..Perhaps I removed this line by accident.

packages/@lwc/jest-jsdom-custom-test-env/src/__tests__/jsdom-custom-test-env.test.js
nolanlawson: Right, I understand that this is just for validation. But the fact that `__filename` works at all is, to me, evidence that we don't necessarily need a custom environment.

I did find [this SO post](https://stackoverflow.com/a/63002088) mentioning a custom environment as the solution, but in my own testing `__filename` also works.

packages/@lwc/jest-ssr-snaphot-utils/src/ssr-snapshot-utils.js
nolanlawson: I think we can avoid overriding the snapshot content if we use a serializer other than `@lwc/jest-preset/src/ssr/snapshot-serializer.js`. The goal is just to serialize the HTML as-is (or perhaps with pretty-printing), yes?

packages/@lwc/jest-jsdom-custom-test-env/src/__tests__/jsdom-custom-test-env.test.js
lesya7: I see what you mean.I need to get an absolute path to a test file from the utils, not directly from the test. I don't want a developer to pass in '__filename' from the test suite. So, instead, I want to get the path from the utils. If I use "__filename" from utils, I will get the path to the utils. Let me know if that addresses your concern.

packages/@lwc/jest-ssr-snaphot-utils/src/ssr-snapshot-utils.js
lesya7: I'm working on that.

packages/@lwc/jest-preset/ssr/jest-preset.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     // Jest 26 default testEnvironment is jsdom: https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults
     testEnvironment: 'node',
     moduleFileExtensions: ['ts', 'js', 'html'],
-    snapshotSerializers: [require.resolve('../src/ssr/html-serializer.js')],
```
lesya7, line 11: Removed the custom serializer and utilized the out-of-the-box "jest-serializer-html".

packages/@lwc/jest-ssr-snaphot-utils/src/ssr-snapshot-utils.js
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Renders the component's markup, captures it in a snapshot that has a unique snapshot hash.
+ *
+ * @param {string} tagName - The name of the HTML tag or component.
+ * @param {Function} Ctor - The constructor of the Lightning Web Component.
+ * @param {Object} [props={}] - An object representing the properties of the component.
+ * @param {Object} [customTestEnv={}] - An object representing the custom test env where the component is being validated.
+ * @returns {{markup: string, snapshotHash: string}} - An object containing the rendered markup and the generated snapshot hash.
+ */
+function renderAndHashComponent(tagName, Ctor, props = {}, customTestEnv = {}) {
```
lesya7, line 15: Renamed the function name after its scope has changed.

package.json
```diff
@@ -31,9 +31,10 @@
         "husky": "^9.0.11",
         "jest": "^29.7.0",
         "jest-environment-jsdom": "^29.7.0",
+        "jest-serializer-html": "7.1.0",
         "lerna": "^8.1.2",
         "lint-staged": "^15.2.2",
-        "prettier": "^3.2.5",
+        "prettier": "^2.0.0",
```
lesya7, line 37: "prettier" v3. doesn't work with inline snapshots. Here is the thread: https://github.com/jestjs/jest/issues/14305. It's going to be fixed in jest v.30. 
Let me know if you want to keep "prettier" v.3 and I would introduce an alias for "prettier" v.2.

packages/@lwc/jest-jsdom-custom-test-env/package.json
```diff
@@ -0,0 +1,23 @@
...
+        "directory": "packages/@lwc/jest-jsdom-custom-test-env"
+    },
+    "bugs": {
+        "url": "https://github.com/salesforce/lwc-test/issues"
+    },
+    "license": "MIT",
+    "main": "src/jsdom-custom-test-env.js",
+    "peerDependencies": {
+        "jest": "^26 || ^27 || ^28 || ^29",
+        "jest-environment-jsdom": "^28 || ^29"
```
lesya7, line 18: Added "jest-environment-jsdom" as a peer dependency.

packages/@lwc/jest-ssr-snaphot-utils/src/ssr-snapshot-utils.js
lesya7: I removed all previously used custom serializers (the one that was replacing some DOM attributes and the one I introduced.)

packages/@lwc/jest-jsdom-custom-test-env/src/__tests__/jsdom-custom-test-env.test.js
nolanlawson: You cannot get the path from the utils unfortunately. Inside of the utils, `__filename` will resolve to the filename of the utils file, not the file that is calling it.

I agree that asking the developer to pass in `__filename` is ugly. However, compared to a custom environment, I think it's less complex and more maintainable.

packages/@lwc/jest-preset/ssr/jest-preset.js
nolanlawson: This is still not where I'd like to go with this... Let me explain.

There are two SSR serialization use cases:

1. SSR alone (current solution)
2. SSR+hydration (proposed solution)

For #<span></span>1, you actually want `/src/ssr/html-serializer.js`. It removes LWC-specific framework attributes, and plus it's what customers have historically used, so dropping it would be a breaking change.

For #<span></span>2, you just need the raw HTML (or perhaps raw+prettified). So this is where `jest-serializer-html` is ideal.

I think that trying to serve both use cases in the same preset is going to cause problems. Unless we decide to deprecate `src/ssr/html-serializer.js` entirely and release a breaking change, and point people to `jest-serializer-html`. But in that case, I don't want us to just be a wrapper around `jest-serializer-html`, because then we run the risk of breaking people whenever that dependency changes.

Let me think about this.

package.json
nolanlawson: Using prettier v2 is fine.

packages/@lwc/jest-jsdom-custom-test-env/src/__tests__/jsdom-custom-test-env.test.js
lesya7: I checked in with our customer, and he confirmed that he doesn't want to include __filename in every single test. Hence, I had to find a way to compute the path differently. I don't think it's risky as long as all configuration is done in "preset".

packages/@lwc/jest-preset/ssr/jest-preset.js
lesya7: Sure. I have very little context into the existing serializers and their value. My idea was just to capture the formatted SSR-ed markup in SSR workflow. And for the SSR + hydration, just consume it (no need for another custom serializer).

packages/@lwc/jest-jsdom-custom-test-env/src/__tests__/jsdom-custom-test-env.test.js
nolanlawson: I'm sure the customer doesn't want to type extra code... but the maintenance burden has to be considered here. Let's talk about this.

packages/@lwc/jest-preset/src/ssr/html-serializer.js
nolanlawson: It doesn't have to be a breaking change if we propose two separate presets.

jest.config.js
```diff
@@ -26,20 +26,20 @@ function integration({ nativeShadow }) {
     return {
         displayName: {
             name: `integration (${nativeShadow ? 'native' : 'synthetic'} shadow)`,
-            color: nativeShadow ? 'blue' : 'cyan'
+            color: nativeShadow ? 'blue' : 'cyan',
         },
 
         rootDir: '<rootDir>/test',
-        preset: '@lwc/jest-preset',
```
nolanlawson, line 33: I would recommend calling it something like `ssr-for-hydration` since otherwise `ssr-client` may be a bit confusing.

jest.config.js
```diff
@@ -53,14 +53,14 @@ module.exports = {
         {
             displayName: {
                 name: `integration (ssr)`,
-                color: 'magenta'
+                color: 'magenta',
             },
 
             rootDir: '<rootDir>/test',
-            preset: '@lwc/jest-preset/ssr',
```
nolanlawson, line 60: This is a breaking change if we're renaming this. Whereas if we avoid renaming it, then it's not a breaking change for existing consumers.

packages/@lwc/jest-preset/ssr-server/jest-preset.js
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+const originalSSRPreset = require('../ssr/jest-preset.js');
+
+module.exports = {
+    ...originalSSRPreset,
+    snapshotSerializers: ['jest-serializer-html'],
```
nolanlawson, line 11: If we make `jest-serializer-html` a dependency and include it in this preset, then we are responsible for breaking changes in `jest-serializer-html`.

Given that this package just pretty-prints the HTML... I dunno. If it were me, I would probably just use our own serializer that keeps the HTML as-is. For instance, the existing serializer could have an option to disable the attribute stripping.

I know I mentioned that pretty-printing may be nice, but it's kind of a hassle if we have to distribute a package. Your call though.

nolanlawson (changes requested):
See above comments

jest.config.js
lesya7: Could you provide more details? I believe the jest.config.js file is being used only for the tests in the "test" folder and isn't directly utilized in the customers' source code. Changing the preset should be fine as long as the tests still pass.

packages/@lwc/jest-preset/ssr-server/jest-preset.js
lesya7: I'm satisfied with the current solution. If any new requirements come up, we can certainly revisit it.

lesya7: Also, could you elaborate on the scope of existing serializers? Why don't we consume the generated markup as-is?

jest.config.js
lesya7: In the next release, we could inform customers that starting from this version, it's recommended they use the latest version.

jest.config.js
lesya7: Sure.

packages/@lwc/jest-preset/ssr-server/jest-preset.js
nolanlawson: That's a great question. I think originally the goal was to have a very lightweight HTML formatter. I wouldn't be surprised if, when we first wrote, there were no good off-the-shelf options (especially for shadow DOM, which didn't even have an HTML format until [declarative shadow DOM](https://web.dev/articles/declarative-shadow-dom) was standardized.

Meanwhile, we also thought it would be useful to hide framework-internal details to avoid breaking snapshots when the framework changes... but this value is debatable IMO.

packages/@lwc/jest-jsdom-custom-test-env/package.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "@lwc/jest-jsdom-custom-test-env",
```
nolanlawson, line 2: The name `custom` does not really communicate much here. I'd call it `@lwc/jest-environment-for-hydration` (to make it clear that it pairs with the other `-for-hydration` one).

nolanlawson (changes requested):
I have one last nitpick, then :+1:

nolanlawson (approved):

PR #300, opened 2024-09-05
@W-16676971: drop support for Jest v.26 and v.27

lesya7:


packages/@lwc/jest-transformer/src/index.js
```diff
@@ -169,8 +169,6 @@ module.exports = {
             transformConfig = rest[0];
             configString = transformConfig.configString;
         } else if (rest.length === 2) {
-            // Handle jest@26 arguments
-            // type getCacheKey = (sourceText: string, sourcePath: string, configStr: string, options: any) => string;
```
nolanlawson, line 173: It looks like the whole `else if` block can be removed.

.circleci/config.yml
```diff
@@ -62,12 +62,6 @@ jobs:
                 command: |
                   yarn remove jest jest-environment-jsdom -W --dev
                   yarn add jest@<<parameters.jest-version>> -W --dev
-            - when:
-                condition:
-                  equal: [<<parameters.jest-version>>, "26"]
-                steps:
-                  - run: yarn clean
-                  - run: yarn install
```
nolanlawson, line 70: The condition below is no longer needed since jest 28/29 is now implied (if the jest version is not "local"):

```yaml
            - when:
                condition:
                  or:
                    - equal: [<<parameters.jest-version>>, "28"]
                    - equal: [<<parameters.jest-version>>, "29"]
```

packages/@lwc/jest-transformer/src/index.js
lesya7: There is some comment regarding v.27 https://github.com/salesforce/lwc-test/pull/300/files#diff-f1a17bcd2dc683c79265f4962796d0f56d3b8c8771968b2256ae6483daad547bR167. Do I need just to remove it?

nolanlawson (approved):

packages/@lwc/jest-transformer/src/index.js
nolanlawson: Sorry, I missed this. The comment is harmless but yes I guess it is redundant now.

lesya7: I'll remove it .

PR #315, opened 2024-09-26
@W-16845391 Revert the breaking change for jest.fn()

Shinoni:
https://github.com/salesforce/lwc-test/pull/260


abhagta-sfdc (approved):

nolanlawson:
FYI @nwcm – we seem to have ended up with breaking changes related to #260 so we are reverting for now. The breakage seems to arise from folks who are already using mocks. Sorry about that.

nolanlawson:
@nwcm Investigation is still ongoing, we may be able to provide details when we get to the bottom of it. Reverting proactively for now.

PR #339, opened 2024-11-15
fix(snapshots): ignore `data-lwc-host-scope-token`

nolanlawson:
We recently added a new framework-specific HTML attribute in https://github.com/salesforce/lwc/pull/4865. This should be ignored by the Jest snapshotter so that it doesn't pollute the snapshot output with attributes that component authors don't care about.

### DIFF
--- a/packages/@lwc/jest-preset/src/ssr/html-serializer.js
+++ b/packages/@lwc/jest-preset/src/ssr/html-serializer.js
@@ -108,6 +108,7 @@ function formatHTML(src) {
             // `data-lwc-host-mutated` may or may not have an attribute value depending on the version of LWC.
             // See: https://github.com/salesforce/lwc/pull/4385
             .replace(/ data-lwc-host-mutated(="[^"]*")?/g, '')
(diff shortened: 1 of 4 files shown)


packages/@lwc/jest-preset/src/ssr/html-serializer.js
```diff
@@ -108,6 +108,7 @@ function formatHTML(src) {
             // `data-lwc-host-mutated` may or may not have an attribute value depending on the version of LWC.
             // See: https://github.com/salesforce/lwc/pull/4385
             .replace(/ data-lwc-host-mutated(="[^"]*")?/g, '')
+            .replace(/ data-lwc-host-scope-token(="[^"]*")?/g, '')
```
wjhsf, line 111:
```suggestion
            .replace(/ data-lwc-[a-z-]+(="[^"]*")?/g, '')
```
future proofing?

nolanlawson: Meh. I like being explicit 🤷

PR #397, opened 2025-10-15
@lwc/jest-preset: improve the logging experience

rochejul:
Here a repo to explain and illustrate the `why` of this PR: https://github.com/rochejul/lwc-jest-console-issue
The big picture: LWC profixies objects and if we use the Console api, we could have some confusion in Jest's logs.

This PR introduces a new flag to allow "logging formatting" for LWC tests

### DIFF
--- a/packages/@lwc/jest-preset/src/console.js
+++ b/packages/@lwc/jest-preset/src/console.js
@@ -1,0 +1,16 @@
+const { unwrap } = require('@lwc/engine-dom');
+
+const originalConsole = global.console;
+const augmentedConsole = {
+    ...global.console,
+};
+
+['log', 'info', 'err', 'dir', 'warn', 'debug'].forEach((methodName) => {
+    augmentedConsole[methodName] = (...args) => {
+        originalConsole[methodName].apply(originalConsole, [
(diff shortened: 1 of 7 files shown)


abhagta-sfdc (approved):

packages/@lwc/jest-preset/src/console.js
```diff
@@ -0,0 +1,16 @@
...
+const originalConsole = global.console;
+const augmentedConsole = {
+    ...global.console,
+};
+
+['log', 'info', 'err', 'dir', 'warn', 'debug'].forEach((methodName) => {
+    augmentedConsole[methodName] = (...args) => {
+        originalConsole[methodName].apply(originalConsole, [
+            ...Array.from(args).map((obj) => unwrap(obj)),
+        ]);
```
wjhsf:
```suggestion
        originalConsole[methodName](...args.map(unwrap));
```
1. `args` is already an array, no need for `Array.from`.
2. `unwrap` only has a single parameter, no need to wrap with an arrow function.
3. The result of `.map` is already a new array, no need to clone it.
4. `originalConsole[methodName]()` already has `this` correctly set, no need for `.apply`.

PR #420, opened 2026-04-15
Only allow enablePrivateMethod to be true for lightning & interop namespaces

a-chabot:


test/src/modules/lightning/privateMethods/__tests__/privateMethods.test.js
```diff
@@ -9,7 +9,7 @@ import PrivateMethods from '../privateMethods';
 
 describe('private methods', () => {
     it('should allow calling private methods from public methods', () => {
-        const element = createElement('smoke-private-methods', { is: PrivateMethods });
+        const element = createElement('lightning-private-methods', { is: PrivateMethods });
```
wjhsf, line 12: Can we test both `smoke-private-methods` and `lightning-private-methods`, since they'll behave differently?

packages/@lwc/jest-transformer/src/__tests__/extract-namespace.test.js
```diff
@@ -0,0 +1,53 @@
...
+    })),
+}));
+
+jest.mock('@babel/core', () => ({
+    transform: jest.fn(() => ({ code: '/* babel */' })),
+}));
+
+jest.mock('@lwc/template-compiler/dist/index.cjs', () => ({
+    generateScopeTokens: jest.fn(() => ({ cssScopeTokens: undefined })),
+}));
```
wjhsf:
```suggestion
```
None of these are used in the string match helper.

packages/@lwc/jest-transformer/src/__tests__/extract-namespace.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+const { extractNamespace } = require('../index');
+
+describe('extractNamespace', () => {
```
wjhsf, line 3: 1. Are the filepaths always absolute or do we need tests for relative?
2. Add tests for `modules/x/foo/bar.js` and `modules/x/foo/bar/baz.js`.

test/src/modules/lightning/privateMethods/__tests__/privateMethods.test.js
a-chabot: sure!

wjhsf (approved):

test/src/modules/smoke/privateMethods/__tests__/privateMethods.test.js
```diff
@@ -1,31 +1,16 @@
...
+        jest.resetModules();
+        expect(() => {
+            require('../privateMethods');
+        }).toThrow(
+            expect.objectContaining({
+                message: expect.stringMatching(
+                    /LWC1007:[\s\S]*Class private methods are not enabled/
+                ),
+            })
+        );
```
wjhsf:
```suggestion
        }).toThrow(/LWC1007:[\s\S]*Class private methods are not enabled/);
```

```suggestion
        }).toThrowError(/LWC1007:[\s\S]*Class private methods are not enabled/);
```
One or the other is equivalent.

packages/@lwc/jest-transformer/src/extract-namespace.js
```diff
@@ -0,0 +1,32 @@
...
+
+    // Match patterns: modules/{namespace} or jest-modules/{namespace}
+    const match = normalizedPath.match(/(?:^|\/)(modules|jest-modules)\/([^/]+)\//);
+
+    if (match && match[2]) {
+        return match[2];
+    }
+
+    // Fallback to 'x', which was previously the default
+    return 'x';
```
wjhsf:
```suggestion
```
we don't actually _need_ a default, and it's cleaner to return a non-string value to differentiate between `x` detected and nothing detected.

wjhsf (approved):

test/src/modules/smoke/privateMethods/__tests__/privateMethods.test.js
a-chabot:
```
    TypeError: expect(...).toThrowError is not a function
``` 

so I went with the `toThrow` option

PR #435, opened 2026-09-10
@W-23885879 feat(vitest-plugin): Plugin resolve/load mechanism for @salesforce/* imports

pragya0702:
## Summary

Part of the Jest→Vitest migration (story **A2**). Wires `@lwc/vitest-plugin`
to **recognize and claim** LWC's `@salesforce/*` and `@label/` scoped imports
so they can be mocked in Vitest unit tests — the Vitest counterpart to the
Babel transforms in `@lwc/jest-transformer`.

This story delivers the **routing/gate mechanism only**: it claims the correct
specifiers and serves a placeholder module. The real per-type mock values are
deferred to follow-up stories (A3–A10).

## What changed

- **`vitest-plugin/src/index.js`**
  - `SCOPED_IMPORT_PREFIXES` — the set of scoped-import prefixes the plugin owns.
    Mirrors the wired `*_IMPORT_IDENTIFIER` consts in `@lwc/jest-transformer`
    (source of Jest parity).
  - `resolveId` (`enforce: 'pre'`) — claims any specifier matching an owned
    prefix as a virtual `\0salesforce:` module; returns `null` for everything
    else, so real packages (e.g. `@salesforce/wire-service-jest-util`) resolve
    normally.
  - `load` — serves the claimed virtual module (placeholder default export
    until A3–A10 add real values).
- **`vitest-preset/src/index.js`** — adds `server.deps.inline` for
  `@salesforce/*` + `@label/` so these run through Vite's transform pipeline
  instead of esbuild pre-bundling (which would bypass the plugin hooks).
- **`vitest-plugin/src/__tests__/scoped-imports-gate.test.mjs`** — gate test:
  the plugin claims every mocked specifier, passes through everything else, and
  handles edges (apex subsuming apexContinuation, loose apex/schema prefixes,
  the legacy `@label/` prefix, bare `@salesforce/label`).

### DIFF
--- a/packages/@lwc/vitest-plugin/src/index.js
+++ b/packages/@lwc/vitest-plugin/src/index.js
@@ -5,11 +5,49 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-// Vite plugin: mocks `@salesforce/*` scoped imports for LWC unit tests (skeleton).
+// Vite plugin: mocks `@salesforce/*` + `@label/` scoped imports for LWC unit tests.
+// Prefixes mirror the wired Jest transforms' import identifiers (jest-transformer/src/transforms/*) — keep in sync.
+const SCOPED_IMPORT_PREFIXES = [
+    '@salesforce/accessCheck/',
+    '@salesforce/apex', // also claims @salesforce/apexContinuation by prefix
+    '@salesforce/client/',
+    '@salesforce/contentAssetUrl/',
+    '@salesforce/customPermission/',
+    '@salesforce/i18n/',
+    '@salesforce/label/',
+    '@label/',
+    '@salesforce/messageChannel/',
+    '@salesforce/resourceUrl/',
+    '@salesforce/schema',
+    '@salesforce/site/',
+    '@salesforce/user/Id',
+    '@salesforce/user/isGuest',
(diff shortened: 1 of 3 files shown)


hkondetiSF (approved):
LGTM

packages/@lwc/vitest-plugin/src/index.js
```diff
@@ -5,11 +5,49 @@
...
+function isMockedSpecifier(source) {
+    return SCOPED_IMPORT_PREFIXES.some((prefix) => source.startsWith(prefix));
+}
+
 export function salesforceScopedImports() {
     return {
         name: '@lwc/vitest-plugin:salesforce-scoped-imports',
-        // TODO: resolveId/load the @salesforce/* virtual modules per emit shape
+        // Claim the specifier before Vite core resolution and before @lwc/rollup-plugin.
+        enforce: 'pre',
```
shugo111, line 39: Why do we require Pre? Cant we just execute it serially?

pragya0702: It already runs serially, `pre` is added just as a guardrail. It makes our plugin run before Vite's built-in resolver, so our mocks resolve first in case an instance of the package is installed locally (unlikely, but added as a defence against that edge case).

shugo111 (approved):

PR #437, opened 2026-09-22
feat(vitest-plugin): mock @salesforce/site/* scoped import values

pragya0702:
## What

Mocks `@salesforce/site/*` scoped imports in the Vite plugin, porting the values from the Jest transform (`jest-transformer/src/transforms/site-scoped-import.js`):

| Specifier | Mock value |
|---|---|
| `@salesforce/site/Id` | `"005000000000000000"` |
| `@salesforce/site/activeLanguages` | `[{ code: 'en-US', label: 'English (US)' }]` |
| `@salesforce/site/defaultLanguages` | `{ code: 'en-US', label: 'English (US)' }` |

## Why

Part of the Jest→Vitest migration (A7). Components importing `@salesforce/site/*` need these values mocked in unit tests, the same way the Jest transformer provides them today.

## Jest parity

- **Narrow claim.** Jest's site transform is a `switch` with only `Id`/`activeLanguages`/`defaultLanguages` and **no `default`**, so any other `@salesforce/site/*` key is left untouched. This PR mirrors that: only the three known keys are claimed; unknown keys (e.g. `@salesforce/site/foo`) pass through to real resolution.
- **`Object.hasOwn`, not `in`.** The key lookup uses `Object.hasOwn` so inherited prototype members (`toString`, `constructor`, `__proto__`, …) are **not** falsely claimed — matching Jest, which never had a case for them.
- Site was removed from the broad `SCOPED_IMPORT_PREFIXES` placeholder list, since it now has real values (it would otherwise over-claim unknown keys).

## Design note

Values are assigned **inline in `index.js`** (parallel to A3's `getStringValue`). Extracting all value logic into a separate `mocks.js` module (`index.js` = Vite wiring only) is under consideration for the wider migration but intentionally deferred — it's a test-invisible refactor that can land later.

## Testing

- `site-scoped-import.test.mjs` asserts all three values through the real `resolveId → load` pipeline via the shared `makeLoad` helper, plus a pass-through assertion for unknown keys.
- Interim runner is `node:test` (`.mjs`, hidden from Jest + the publish glob) until Vitest is wired up: `node --test src/__tests__/*.test.mjs`.


hkondetiSF (approved):
LGTM — Just flag: shares load() + SCOPED_IMPORT_PREFIXES with the other PRs [436](https://github.com/salesforce/lwc-test/pull/436), so this'll need a rebase once the prior one lands. Please rebase + re-run tests before merge.

